PR #363, opened 2020-05-28
feat(btc): implement `LedgerService`

### LINKED ISSUE #206: Implement `LedgerService` for `BTC`

sleepdefic1t:
## Summary

- add 'got' dep to `platform-sdk-btc`.
- fix client nock url & tests for `platform-sdk-btc`.
- update btc joi schema.
- add ledger btc dependencies.
- implement btc support in 'ledger.ts'.
- add ledger btc fixtures.
- add ledger btc unit tests.
- add btc test ci workflow.

## Checklist

- [x] Tests
- [x] Ready to be merged

## Additional Comments

The implementation is working per tests in `hw-app-btc` for basic legacy/P2PKH transactions, but I'm not certain if this would be correct in the full scope of `platform-sdk`, or where certain aspects should be handled (externally vs. within LedgerService itself).

A more robust and full implementation can be seen here for reference:
[LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/libcore-signOperation.js#L19](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/libcore-signOperation.js#L19)

---

Resolves #206


faustbrian:
> The implementation is working per tests in `hw-app-btc` for basic legacy/P2PKH transactions, but I'm not certain if this would be correct in the full scope of `platform-sdk`, or where certain aspects should be handled (externally vs. within LedgerService itself).
> 
> A more robust and full implementation can be seen here for reference:
> [LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/libcore-signOperation.js#L19](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/libcore-signOperation.js?rgh-link-date=2020-05-28T03%3A33%3A38Z#L19)

For now the implementation is fine as is and we'll see where we have to expand or alter the ledger implementations when we get to a point where we have multiple coins implemented into the wallets.

sleepdefic1t:
@faustbrian 
We good then 👍

PR #381, opened 2020-06-02
feat(intl): implement `DateTime#toDate/startOf/from/fromNow` methods

goga-m:
## Summary
Adds support for additional  date methods (provided by `dayjs`): 

`Datetime#toDate` - To get a copy of the native Date object.
`DateTime#startOf` - Returns cloned Day.js object and set it to the start of a unit of time.
`DateTime#from` - Returns the string of relative time from X (e.g 3 years ago).
`DateTime#fromNow` - Returns the string of relative time from X.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-intl/src/datetime.ts
```diff
@@ -354,6 +356,22 @@ export class DateTime {
 		return this.#instance.valueOf();
 	}
 
+	public toDate(): Date {
```
faustbrian, line 359: This method should be above `valueOf`.

packages/platform-sdk-intl/src/datetime.ts
```diff
@@ -354,6 +356,22 @@ export class DateTime {
...
+		return this.#instance.toDate();
+	}
+
+	public startOf(unit: QUnitType): DateTime {
```
faustbrian, line 363: The `startOf`, `from` and `fromNow` methods should be above the `toObject` method.

PR #414, opened 2020-06-13
chore: update dependencies

faustbrian:


faustbrian:
Ledger has an implicit dependency so that needs to be fixed before this can be merged.

```
@ledgerhq/hw-app-btc tried to access @ledgerhq/logs, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: @ledgerhq/logs (via "@ledgerhq/logs")
Required by: @ledgerhq/hw-app-btc@npm:5.16.0 (via /home/runner/work/platform-sdk/platform-sdk/.yarn/cache/@ledgerhq-hw-app-btc-npm-5.16.0-40feab515a-2.zip/node_modules/@ledgerhq/hw-app-btc/lib/)
```

PR #484, opened 2020-06-25
refactor(profiles): use `Collection` as underlying `DataRepository`

faustbrian:
https://github.com/ecrmnn/collect.js/


faustbrian:
Seems to have some issues with class instances as values. Will close this until I rolled my own implementation.

PR #716, opened 2020-09-24
refactor: adjust return types in notifications repo

goga-m:
## Summary
- Return array of `NotificationsWithId` from read/unread methods instead of object
- Use `NotificationsWithId` as return type instead of `Notification`
- Add coverage

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
Is the `NotificationWithId` even still needed or should `Notification` always be returned with an ID?

goga-m:
It was just only needed for push but better with Except<Notification, "id"> :+1:

PR #750, opened 2020-10-16
fix(ark): validate address before checking neo network

dated:
## Summary

Validates an ARK address before querying the NEO network for it to avoid any bad input.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian (changes requested):
Needs a test to ensure that the method wasn't called if an invalid address is used.

PR #779, opened 2020-11-16
refactor: fetch exchange rates only for live coins

dated:
## Summary

Right now the wallet fetches the exchange rate for all coins, live or not. With these change the respective methods return early instead of fetching the rate for non-live coins.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
+++ b/packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
@@ -26,9 +26,9 @@ export class ExchangeRateService {
 			wallets = profile
 				.wallets()
 				.values()
-				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency && !wallet.network().isTest());
+				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency && wallet.network().isLive());
 		} else {
-			wallets = wallets.filter((wallet: ReadWriteWallet) => !wallet.network().isTest());
+			wallets = wallets.filter((wallet: ReadWriteWallet) => wallet.network().isLive());
 		}
 
 		if (!wallets.length) {
--- a/packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
+++ b/packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
@@ -29,6 +29,10 @@ export class ExchangeRateService {
 				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency);
 		}
 
+		if (!wallets.length || !wallets[0].network().isLive()) {
+			return;
+		}
+
 		const marketService = MarketService.make(
 			profile.settings().get(ProfileSetting.MarketProvider) || "coingecko",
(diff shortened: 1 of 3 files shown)


packages/platform-sdk-profiles/src/wallets/wallet.ts
```diff
@@ -169,7 +169,7 @@ export class Wallet implements ReadWriteWallet {
 	}
 
 	public convertedBalance(): BigNumber {
-		if (this.network().isTest()) {
+		if (!this.network().isLive()) {
```
faustbrian, line 172: Why is this necessary? If the `isTest` check doesn't work then `isTest` is bugged.

packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
```diff
@@ -29,6 +29,10 @@ export class ExchangeRateService {
 				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency);
 		}
 
+		if (!wallets.length || !wallets[0].network().isLive()) {
```
faustbrian, line 32: Can't we already filter out `isTest` wallets before we even filter the currency? Then the second filter would be a noop if there are no live wallets for that currency.

packages/platform-sdk-profiles/src/wallets/wallet.ts
dated: not necessary at all, changing back.

packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
```diff
@@ -26,7 +26,13 @@ export class ExchangeRateService {
 			wallets = profile
 				.wallets()
 				.values()
-				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency);
+				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency && !wallet.network().isTest());
```
faustbrian, line 29:
```suggestion
				.filter((wallet: ReadWriteWallet) => wallet.currency() === currency && wallet.network().isLive());
```

packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
```diff
@@ -26,7 +26,13 @@ export class ExchangeRateService {
...
+		} else {
+			wallets = wallets.filter((wallet: ReadWriteWallet) => !wallet.network().isTest());
```
faustbrian, line 31:
```suggestion
			wallets = wallets.filter((wallet: ReadWriteWallet) => wallet.network().isLive());
```

packages/platform-sdk-profiles/src/environment/services/exchange-rate-service.ts
dated: Right now the methods for the exchange rate are based on `allByCoin`, whose purpose is questionable at this point, as it groups live & test wallets together.

PR #792, opened 2020-11-30
test: increase unit coverage for `WalletData`

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian (changes requested):
Keep the tests separate instead of in those `Implemented` and `Not Implemented` blocks. Makes it easier to read and modify them without having to figure out what is looping over what to call what.

PR #857, opened 2020-12-09
test: add coverage for `src/environment`

goga-m:
## Summary
Depends on https://github.com/ArkEcosystem/platform-sdk/pull/856 as the `usedCoinsWithNetworks` method is now obsolete and coverage is unecessary.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/environment/env.ts
+++ b/packages/platform-sdk-profiles/src/environment/env.ts
@@ -34,6 +34,7 @@ export class Environment {
 		container.set(Identifiers.FeeService, new FeeService());
 		container.set(Identifiers.KnownWalletService, new KnownWalletService());
 		container.set(Identifiers.WalletService, new WalletService());
+		container.set(Identifiers.Migrator, new Migrator());
 
 		container.set(Identifiers.Coins, options.coins);
 	}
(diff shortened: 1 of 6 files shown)


packages/platform-sdk-profiles/src/environment/env.ts
```diff
@@ -164,11 +159,11 @@ export class Environment {
 			throw new Error("Please call [verify] before booting the environment.");
 		}
 
-		if (this.storage.data) {
+		if (this.storage.data && Object.keys(this.storage.data).length > 0) {
 			this.data().fill(this.storage.data);
 		}
 
-		if (this.storage.profiles) {
+		if (this.storage.profiles && Object.keys(this.storage.profiles).length > 0) {
```
goga-m, line 166: @faustbrian added additional condition here as `storage.data` and `storage.profiles` are always set as objects from `verify`

packages/platform-sdk-profiles/src/environment/env.ts
```diff
@@ -136,13 +137,7 @@ export class Environment {
...
 
-		const validated = schema.validateSync(
-			{ data, profiles },
-			{
-				strict: true,
-			},
-		);
-
+		const validated = validator.validate({ data, profiles }, schema, { strict: true });
 		if (validator.fails()) {
```
goga-m, line 141: @faustbrian also moved this validation into `Validator` as the `validation as `validator.fails()` was always false previously with `schema.validateSync`

packages/platform-sdk-profiles/src/environment/env.ts
faustbrian: The condition before the `&&` is probably not needed then since it will always be defined.

faustbrian:
Linting is failing @goga-m

PR #858, opened 2020-12-09
test(profiles): add coverage for `src/profiles`

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/profiles/profile.ts
+++ b/packages/platform-sdk-profiles/src/profiles/profile.ts
@@ -242,6 +242,7 @@ export class Profile implements ProfileContract {
 	 */
 	public async restore(password?: string): Promise<void> {
 		let data: ProfileStruct | undefined;
+		let errorReason = "";
 
 		try {
 			if (password !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/platform-sdk-profiles/src/profiles/profile.ts
```diff
@@ -381,8 +383,8 @@ export class Profile implements ProfileContract {
 	private decrypt(password: string): ProfileStruct {
 		const usesPassword: boolean = this.#data.password !== undefined;
 
-		if (usesPassword && password === undefined) {
-			throw new Error("This profile uses a password but none was passed for decryption.");
+		if (!usesPassword) {
+			throw new Error("This profile does not use a password but password was passed for decryption");
```
goga-m, line 387: @faustbrian this condition was unreachable because `restore` ensures password is always defined. Modified the condition to throw when password is provided for a non-password profile restore.

packages/platform-sdk-profiles/src/profiles/profile.ts
```diff
@@ -242,19 +242,21 @@ export class Profile implements ProfileContract {
...
-		} catch {
-			// Do nothing if both fail...
+		} catch (e) {
+			// Append the error reason
+			errorReason = ` Reason: ${e.message}`;
 		}
 
 		if (data === undefined) {
-			throw new Error("Failed to decode or decrypt the profile.");
+			throw new Error(`Failed to decode or decrypt the profile.${errorReason}`);
```
goga-m, line 259: @faustbrian there could be cases that the reason might be helpful, e.g restoring and decrypting a passwordless profile would error that the password is not required. https://github.com/ArkEcosystem/platform-sdk/pull/858/files#diff-f4f13df0c990dcf1ed7c0271688dfb420d39dbc82a611b044bfc63414b2fb01dR284

PR #862, opened 2020-12-11
refactor(profiles): apply migrations per-profile instead of per-environment

faustbrian:


faustbrian:
Task linked: [CU-bazpty Applying migrations to encrypted profiles](https://app.clickup.com/t/bazpty)

PR #870, opened 2020-12-16
fix(profiles): set in-memory password when setting the password

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-profiles/src/profiles/authenticator.ts
```diff
@@ -11,6 +12,8 @@ export class Authenticator {
 
 	public setPassword(password: string): void {
 		this.#profile.settings().set(ProfileSetting.Password, Bcrypt.hash(password));
+
+		MemoryPassword.set(this.#profile, password);
```
faustbrian, line 16: Needs a test to ensure that the in-memory password has been set when this method is called.

goga-m: Modified those in `env.test.ts` and `profile-repository` to check this, but I 'll add a separate one for authenticator

PR #874, opened 2020-12-17
refactor(profiles): migrate container to use inversify

faustbrian:
Replaces the naive container with https://inversify.io/ to have proper error handling and singletons.


faustbrian:
Disregarding this in favour of a simpler solution.

PR #890, opened 2021-01-07
refactor(profiles): flush only the aggregate transaction method

luciorubeens:
## Summary

https://app.clickup.com/t/c4zdzu

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-profiles/src/profiles/aggregates/transaction-aggregate.ts
```diff
@@ -39,7 +39,11 @@ export class TransactionAggregate {
 			.includes(true);
 	}
 
-	public flush(): void {
+	public flush(method: string): void {
```
faustbrian, line 42: Make this an optional parameter and either flush all history or a specific history depending on if it is provided or not.

PR #903, opened 2021-01-21
feat(profiles): export plugin logo from metadata

luciorubeens:
## Summary

Get the plugin logo and the full package object.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-profiles/src/plugins/plugin-registry.models.ts
```diff
@@ -107,6 +107,10 @@ export class RegistryPlugin {
 		return null;
 	}
 
+	public logo(): string[] {
```
faustbrian, line 110: Shouldn't this be just `string` since a logo has to be singular?

packages/platform-sdk-profiles/src/plugins/plugin-registry.models.ts
```diff
@@ -131,6 +135,10 @@ export class RegistryPlugin {
 		return Object.keys(this.#data.versions);
 	}
 
+	public raw() {
```
faustbrian, line 138: Since this just aliases `getLatestVersion` you could make `getLatestVersion` public and add a proper type hint for it instead of any.

PR #993, opened 2021-02-09
feat(ada): implement transaction listing

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -27,15 +28,24 @@ export class ClientService implements Contracts.ClientService {
 	}
 
 	public async __destruct(): Promise<void> {
-		//
+		throw new Error("Method not implemented.");
```
faustbrian, line 31: This is always called and is allowed to be empty and do nothing. Some coins need this to terminate their websocket connections.

packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -0,0 +1,205 @@
...
+	public confirmations(): BigNumber {
+		return BigNumber.make(this.data.confirmations);
+	}
+
+	public sender(): string {
+		throw new Exceptions.NotImplemented(this.constructor.name, "sender");
+	}
+
+	public recipient(): string {
+		throw new Exceptions.NotImplemented(this.constructor.name, "recipient");
```
faustbrian, line 27: This would cause the consumer of the package to always crash if they want to list transactions.

packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -0,0 +1,205 @@
...
+	public timestamp(): DateTime | undefined {
+		return DateTime.make(this.data.inserted_at.time);
+	}
+
+	public confirmations(): BigNumber {
+		return BigNumber.make(this.data.confirmations);
+	}
+
+	public sender(): string {
+		throw new Exceptions.NotImplemented(this.constructor.name, "sender");
```
faustbrian, line 23: This would cause the consumer of the package to always crash if they want to list transactions.

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -80,9 +90,11 @@ export class ClientService implements Contracts.ClientService {
 		throw new Exceptions.NotImplemented(this.constructor.name, "broadcastSpread");
 	}
 
-	private async get(path: string, query?: Contracts.KeyValuePair): Promise<Contracts.KeyValuePair> {
-		const response = await this.#http.get(`${this.#peer}/${path}`, query);
+	entityHistory(id: string, query?: Record<string, any>): Promise<Coins.TransactionDataCollection> {
```
faustbrian, line 93: This method can be removed.

packages/platform-sdk-ada/src/dto/transaction.ts
marianogoldman: @faustbrian I need a bit of help here... Given this transaction, what would you use as sender / recipient? For e.g. `inputs.address`? And for output, we have 2, one for the money sent someone else, and another for the change staying back with us. What would be the way to identify out of those 2 which is which?
```
  {
	"inserted_at": {
	  "height": {
		"quantity": 2292394,
		"unit": "block"
	  },
	  "time": "2021-02-05T15:04:16Z",
	  "epoch_number": 112,
	  "absolute_slot_number": 18168240,
	  "slot_number": 153840
	},
	"status": "in_ledger",
	"withdrawals": [],
	"amount": {
	  "quantity": 25168801,
	  "unit": "lovelace"
	},
	"inputs": [
	  {
		"amount": {
		  "quantity": 30000000,
		  "unit": "lovelace"
		},
		"address": "addr_test1qrhvwtn8sa3duzkm93v5kjjxlv5lvg67j530wyeumngu23lk8ttq8f3gag0h89aepvx3xf69g0l9pf80tqv7cve0l33s4s8xvh",
		"id": "6bf76f4380da8a389ae0a7ecccf1922b74ae11d773ba8b1b761d84a1b4474a4f",
		"assets": [],
		"index": 0
	  }
	],
	"direction": "outgoing",
	"fee": {
	  "quantity": 168801,
	  "unit": "lovelace"
	},
	"outputs": [
	  {
		"amount": {
		  "quantity": 25000000,
		  "unit": "lovelace"
		},
		"address": "addr_test1qzct2hsralem3fqn8fupu90v3jkelpg4rfp4zqx06zgevpachk6az8jcydma5a6vgsuw5c37v0c8j6rlclpqajn2vxsq3rz4th",
		"assets": []
	  },
	  {
		"amount": {
		  "quantity": 4831199,
		  "unit": "lovelace"
		},
		"address": "addr_test1qzfjfm724nv9qz6nfyagmj0j2uppr35gzv5qee8s7489wxlk8ttq8f3gag0h89aepvx3xf69g0l9pf80tqv7cve0l33scc4thv",
		"assets": []
	  }
	],
	"metadata": null,
	"depth": {
	  "quantity": 9216,
	  "unit": "block"
	},
	"id": "35b40547f04963d3b41478fc27038948d74718802c486d9125f1884d8c83a31d",
	"deposit": {
	  "quantity": 0,
	  "unit": "lovelace"
	},
	"mint": []
  },
```

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -35,7 +36,16 @@ export class ClientService implements Contracts.ClientService {
 	}
 
 	public async transactions(query: Contracts.ClientTransactionsInput): Promise<Coins.TransactionDataCollection> {
-		throw new Exceptions.NotImplemented(this.constructor.name, "transactions");
+		const transactions: object[] = (await this.get(`v2/wallets/${query.address}/transactions`)) as object[];
```
marianogoldman, line 39: @faustbrian I've got another question here. The `cardano-wallet` api always requires a `walletId`. What field should I use out of those in the contract `ClientTransactionsInput`? I used address, but it's not right. Actually none of them seem to be right.

packages/platform-sdk-ada/src/dto/transaction.ts
faustbrian: I would use the address of the wallet that we are looking up transactions for as the sender and I would ignore the change address and use the actual wallet that received the money.

packages/platform-sdk-ada/src/services/client.ts
faustbrian: What's wrong with the `address` property?

marianogoldman: Well, technically it's not an address. Address look different, they start with `addr_`. This is a "walletId", I think it's something that could be derived from the root key pair.

faustbrian: Do we get this ID from the API when we request the wallet information? If so then we could store this property as meta data on the wallet instance and retrieve that when needed.

sleepdefic1t: Yeah..
Wallet Id is from the Wallet API.. different than address.
I never could track down exactly how it’s derived.

marianogoldman: To the wallet api we submit only our public key, and it responds with the wallet id. So it must be derivable from there. I'm trying to figure this out.
```
POST http://localhost:8090/v2/wallets
Content-Type: application/json

{
  "name": "Shelley publicKey-only Wallet",
  "account_public_key": "52f79be32364ecd0846652a749a876e4eb550889273404991a92007ac88773f2a21d47deda83a49cb435a2adfd0fb6945efc5b07f384db06d27c69cc4f586805",
  "address_pool_gap": 20
}
```
And we get the reponse:
```
{
  "address_pool_gap": 20,
  "state": {
    "status": "syncing",
    "progress": {
      "quantity": 0,
      "unit": "percent"
    }
  },
  "balance": {
    "reward": {
      "quantity": 0,
      "unit": "lovelace"
    },
    "total": {
      "quantity": 0,
      "unit": "lovelace"
    },
    "available": {
      "quantity": 0,
      "unit": "lovelace"
    }
  },
  "name": "Shelley publicKey-only Wallet",
  "delegation": {
    "next": [],
    "active": {
      "status": "not_delegating"
    }
  },
  "id": "82d63879d8914699da73edbb65b091da16e21964",
  "tip": {
    "height": {
      "quantity": 0,
      "unit": "block"
    },
    "time": "2019-07-24T20:20:16Z",
    "epoch_number": 0,
    "absolute_slot_number": 0,
    "slot_number": 0
  },
  "assets": {
    "total": [],
    "available": []
  }
}
```

marianogoldman: Screenshot from 2021-02-10 08-31-26
@faustbrian This is a screenshot from Daedalus. This is how they show transactions. They have senders addresses and receipients addresses. Both can be just one or can be many. How about we use the first of the senders as `sender()`, the first of recipients as `recipient()` and then all recipients as `recipients()`? This way we would only probably miss showing some of the extra senders, but we still show one address that belongs to the sender. If there are more, they should still belong to the same account.

faustbrian: That should work. We will link to the explorer anyways for full details.

packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -11,12 +11,12 @@ export class TransactionData extends DTO.AbstractTransactionData implements Cont
...
 	}
 
-	public timestamp(): DateTime | undefined {
+	public timestamp(): DateTime {
 		return DateTime.make(this.data.inserted_at.time);
 	}
 
 	public confirmations(): BigNumber {
-		return BigNumber.make(this.data.confirmations);
+		return BigNumber.make(NaN);
```
faustbrian, line 19:
```suggestion
		return BigNumber.ZERO;
```

PR #1002, opened 2021-02-10
feat(ada): implement `ClientService#transaction`

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -32,11 +32,16 @@ export class ClientService implements Contracts.ClientService {
 	}
 
 	public async transaction(id: string): Promise<Contracts.TransactionDataType> {
-		throw new Exceptions.NotImplemented(this.constructor.name, "transaction");
+		const walletId = `98c83431e94407bc0889e09953461fe5cecfdf18`; // FIXME
```
marianogoldman, line 35: @faustbrian Where could I get this wallet id from? This is what we were discussing yesterday. I think it should be stored locally in the wallet metadata (profile?) once we create the Cardano wallet. For that, we send the account public key and we get back that id. Any chance we have any place for storing it and using it here?

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -32,11 +32,16 @@ export class ClientService implements Contracts.ClientService {
...
+		const data: object = (await this.get(
+			`v2/wallets/${walletId}/transactions/${id}`,
+		)) as object[];
+		return new TransactionData(data);
 	}
 
 	public async transactions(query: Contracts.ClientTransactionsInput): Promise<Coins.TransactionDataCollection> {
-		const transactions: object[] = (await this.get(`v2/wallets/${query.address}/transactions`)) as object[];
+		const walletId = query.address; // FIXME
```
marianogoldman, line 43: Here we did this, although I didn't like it. But we should take the same approach as for the other endpoint.

packages/platform-sdk-ada/src/services/client.ts
faustbrian: That ID would be handled outside of this package so I'll make a separate task for this. Is it impossible to get the transaction without also knowing the wallet ID?

marianogoldman: The endpoint requires both, transaction is a 'subresource' of wallet.
`https://localhost:8090/v2/wallets/{walletId}/transactions/{transactionId}`

faustbrian: I'll take a look at this tomorrow to see if there is a way to get the transaction without knowing the wallet ID because this method should work without any input except the transaction ID.

faustbrian: Doesn't look like there is a way so I'll leave this as a draft until I figure out how to restructure the contracts to allow for optional parameters like the wallet ID because other coins probably will also have some weird things like that.

PR #1014, opened 2021-02-12
feat(ada): migration from cardano-rest to cardano-graphql

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -69,10 +94,12 @@ export class TransactionData extends DTO.AbstractTransactionData implements Cont
 	}
 
 	public isSent(): boolean {
+		// FIXME Need to find a way to determine this
 		return this.data.direction === "outgoing";
 	}
 
 	public isReceived(): boolean {
+		// FIXME Need to find a way to determine this
```
faustbrian, line 102: We could probably check if the outputs contain one of our addresses.

packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -69,10 +94,12 @@ export class TransactionData extends DTO.AbstractTransactionData implements Cont
 	}
 
 	public isSent(): boolean {
+		// FIXME Need to find a way to determine this
```
faustbrian, line 97: We could probably check if the inputs contain one of our addresses.

packages/platform-sdk-ada/src/dto/transaction.ts
```diff
@@ -46,18 +51,29 @@ export class TransactionData extends DTO.AbstractTransactionData implements Cont
...
 			(output: Contracts.KeyValuePair) =>
 				new DTO.UnspentTransactionData({
-					amount: BigNumber.make(output.amount.quantity),
+					amount: BigNumber.make(output.value),
 					addresses: [output.address],
 				}),
 		);
 	}
 
 	public amount(): BigNumber {
```
faustbrian, line 60: Could we use the `totalOutput` property for this? It seems to exclude the `fee` but I would need to check a transaction that contained changed to see if that is also taken into account.

faustbrian:
Merging this to work on some other things. Remaining changes can be submitted through another PR.

packages/platform-sdk-ada/src/dto/transaction.ts
marianogoldman: @faustbrian Change is also part of the output. That's why I had to do that ridiculous assumption (that change will be the last output address) and the the transferred amount are all the others.

faustbrian: Will have a look at some other possible solutions then.

PR #1027, opened 2021-02-15
feat(ada): wallet balance using graphql api

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-ada/src/services/client.test.ts
```diff
@@ -19,14 +19,18 @@ beforeAll(() => nock.disableNetConnect());
...
-			.get("/v2/wallets/98c83431e94407bc0889e09953461fe5cecfdf18")
-			.reply(200, require(`${__dirname}/../../test/fixtures/client/wallet.json`));
+			.post("/")
+			.reply(200, require(`${__dirname}/../../test/fixtures/client/transactions-0.json`))
+			.post("/")
+			.reply(200, require(`${__dirname}/../../test/fixtures/client/transactions-20.json`))
+			.post("/")
+			.reply(200, require(`${__dirname}/../../test/fixtures/client/utxos-aggregate.json`));
 
-		const result = await subject.wallet("98c83431e94407bc0889e09953461fe5cecfdf18");
```
marianogoldman, line 25: @faustbrian We probably need to define what we consider a wallet id. `98c83431e94407bc0889e09953461fe5cecfdf18` was the one that cardano-wallet returns that we couldn't figure out how to calculate.
So I changed this for using the public key of the account (not sure it's the best idea, but I had to choose something).

faustbrian: Can we use the address? That's what we usually pass in to this method.

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -85,7 +91,37 @@ export class ClientService implements Contracts.ClientService {
...
+			spendAddresses
+				.filter((sa) => usedAddresses.find((ua) => ua === sa) !== undefined)
+				.forEach((sa) => usedSpendAddresses.add(sa));
+			changeAddresses
+				.filter((sa) => usedAddresses.find((ua) => ua === sa) !== undefined)
+				.forEach((sa) => usedChangeAddresses.add(sa));
+
+			exhausted = usedAddresses.length === 0;
+			offset += 20;
+		} while (!exhausted);
```
marianogoldman, line 115: I know this iteration looks ridiculous, but I think it's necessary. Doing multiple queries seem to be the only way to find out which addresses are used and which aren't. Once we have that (after multiple iterations) we can fetch the utxos and by aggregating we calculate the account balance.

packages/platform-sdk-ada/src/services/client.test.ts
marianogoldman: Which address? There's no such thing as an address (that I know of)

faustbrian: I'll make some changes tomorrow to abstract the knowledge about the specific input.

marianogoldman: https://docs.cardano.org/projects/cardano-wallet/en/latest/About-Address-Derivation.html#hd-sequential-wallets-a-la-bip-44

marianogoldman: Ok. Meanwhile I'll do the reminder (transactions, which should be a subcase of what I already did).

faustbrian: Will have to think a bit about how to abstract away the specifics about the input because we allow people to import their wallets by an address in the desktop wallet, which won't be possible in this case.

marianogoldman: My guess is for Cardano users, the wallet id provided by `cardano-wallet` could be something they're used to. Perhaps not for end users because, at least in Daedalus, it's never shown. They just import a passphrase and the wallets are imported.

PR #1034, opened 2021-02-16
feat(ada): broadcast transaction using graphql

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-ada/src/services/transaction.ts
```diff
@@ -61,9 +61,10 @@ export class TransactionService implements Contracts.TransactionService {
 				),
 				createValue(utxos[i].value),
 			);
+			break;
```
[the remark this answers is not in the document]
faustbrian: A witness is usually what we call a signer or participant with our products. I'll have a look later to see what's wrong but seems weird that this error pops up if we are using the correct private key.

faustbrian:
Took the address list parts from this PR and moved them out so that we can have a fixed structure for all coins that need to generate lists of addresses and also need it for https://github.com/ArkEcosystem/platform-sdk/pull/1149.

- https://github.com/ArkEcosystem/platform-sdk/pull/1150
- https://github.com/ArkEcosystem/platform-sdk/pull/1151

packages/platform-sdk-ada/src/services/transaction.ts
```diff
@@ -44,26 +47,41 @@ export class TransactionService implements Contracts.TransactionService {
...
-		const privateKey = getCip1852Account(
-			input.sign.mnemonic,
-			this.#config.get<number>("network.crypto.slip44"),
-		).to_raw_key();
+		const accountKey: Bip32PrivateKey = deriveAccountKey(deriveRootKey(input.sign.mnemonic), 0);
+		const publicKey = accountKey.to_public();
 
-		// These are the inputs (UTXO) that will be consumed to satisfy the outputs. Any change will be transfered back to the sender
-		const utxos: any = await this.listUnspentTransactions(input.from);
+		/* TODO: We need to determine how to specify the input.
```
marianogoldman, line 53: @faustbrian Need your input here.

packages/platform-sdk-ada/src/services/transaction.ts
```diff
@@ -44,26 +47,41 @@ export class TransactionService implements Contracts.TransactionService {
...
+			 both internal (change) and external (spend).
+		 */
+		const utxos: UnspentTransaction[] = await this.listUnspentTransactions(input.from); // when more that one utxo, they seem to be ordered by amount descending
 
+		// Figure out the utxos to use
+		// TODO Need to make sure it covers the fess also. We need to be clever here.
+		const usedUtxos: UnspentTransaction[] = [];
+		const requestedAmount: number = parseInt(input.data.amount); // TODO see if we need to use bigint here
+		let amount: number = 0;
 		for (let i = 0; i < utxos.length; i++) {
```
marianogoldman, line 67: @faustbrian We need to do better than this!!! Let's suppose we have several utxos, we need to come up with some way to pick the best set that covers the amount to transfer + fees.
I saw some wallets offer to use oldest received first, minimize fees, etc. Not sure how we want to approach the utxo selection. How is it for Bitcoin.

packages/platform-sdk-ada/src/services/transaction.ts
```diff
@@ -204,13 +246,24 @@ export class TransactionService implements Contracts.TransactionService {
...
+						hash
 				  }
 				  value
 				}
 			  }`,
 			)
 		).utxos;
 	}
+
+	private async deriveAddressesAndSigningKeys(publicKey: Bip32PublicKey, networkId, accountKey: Bip32PrivateKey) {
```
marianogoldman, line 260: We'll need to figure out how many we need to calculate. In general they suggest using a page size of 20, and then stop when all 20 are unused. But for that we need to issue graphql requests. Which I think is the right thing to do.

faustbrian: Don't think we have any other option so go for it.

packages/platform-sdk-ada/src/services/transaction.ts
faustbrian: I think we would need to use the extended public key here which we should add as a new input property so the consumer can explicitly state that they want the package to perform some kind of derivation to gather all data necessary.

packages/platform-sdk-ada/src/services/transaction.ts
faustbrian: I'll check yoroi and see how they select the UTXO that are consumed.

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -97,10 +72,11 @@ export class ClientService implements Contracts.ClientService {
 	}
 
 	public async wallet(id: string): Promise<Contracts.WalletData> {
-		const { usedSpendAddresses, usedChangeAddresses } = await this.usedAddressesForAccount(id);
+		const { usedSpendAddresses, usedChangeAddresses } = await usedAddressesForAccount(this.#config, id);
```
faustbrian, line 75: Think we should swap the argument here so it matches `fetchTransaction(id, this.#config)`.

PR #1037, opened 2021-02-17
feat(profiles): implement plugin whitelist

goga-m:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/plugins/plugin-repository.ts
+++ b/packages/platform-sdk-profiles/src/plugins/plugin-repository.ts
@@ -14,6 +14,7 @@ export class PluginRepository {
 	readonly #data: DataRepository;
 	readonly #registry: PluginRegistry;
 	readonly #blacklist: Set<string> = new Set<string>();
+	readonly #whitelist: Set<string> = new Set<string>();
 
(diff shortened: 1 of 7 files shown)


packages/platform-sdk-profiles/src/plugins/plugin-repository.ts
```diff
@@ -44,12 +45,24 @@ export class PluginRepository {
...
-	public fill({ data, blacklist }: { data: object; blacklist: string[] }): void {
+	public fill({
+		data,
+		blacklist = [],
+		whitelist = [],
+	}: {
+		data: object;
+		blacklist?: string[];
+		whitelist?: string[];
+	}): void {
```
goga-m, line 56: @faustbrian making them optional (both blacklist and whitelist) that default to empty sets, to prevent profile restoration from throwing "not iterable" error when encrypted/encoded data do not have these set already.

faustbrian:
Task linked: [CU-ck1vwh [Plugins] obey whitelist for listing plugins so we don't list all the crap that's on NPM](https://app.clickup.com/t/ck1vwh)

PR #1039, opened 2021-02-17
refactor(profiles): remove blacklist plugins implementation

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-profiles/src/plugins/plugin-repository.test.ts
```diff
@@ -52,22 +52,11 @@ it("should throw if a plugin cannot be found by its ID", () => {
 });
 
 it("should restore previously created data", () => {
-	subject.fill({ data: { [stubPlugin.id]: stubPlugin }, blacklist: [] });
+	subject.fill({ data: { [stubPlugin.id]: stubPlugin } });
```
faustbrian, line 55: This can now be a flat object, no need for a separate `data` key because we won't store anything else here.

goga-m: Missed it. Thanks

packages/platform-sdk-profiles/src/plugins/plugin-repository.ts
```diff
@@ -74,8 +74,6 @@ export class PluginRepository {
 	}
 
 	public toObject(): object {
```
faustbrian, line 76: This method can be removed and `all` can be called instead to dump it.

PR #1090, opened 2021-02-24
refactor(dot): use SQLite as database engine

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
Task linked: [CU-eh0zk5 Default the DOT indexer to sqlite as storage](https://app.clickup.com/t/eh0zk5)

marianogoldman:
@faustbrian I think it's done, although I have a few things to mention.

1. It's doing exactly what it was doing for elasticsearch but into a sqlite db. But I don't think the data we're persisting is very useful or well digested. For e.g. there's nothing that allows us to query for a transfer transaction. Not sure because I don't know anything about Dot yet but I think we'd need to parse the "method" in the transactions to figure out the transaction type and extract what's useful. I understand you're trying to figure out what's important to index based on what we'll need to query later, so this may get more clear soon.

2. I think there's a pattern that could be extracted for all indexers. For e.g. figure out where to start, fetch blocks + transactions, persist them, repeat until all blocks are consumed. Perhaps we could come up with a template algorithm where each indexer would implement the particulars: fetching transactions (custom for each coin) and requesting storage (elastic search or sqlite for now). Not sure it's worth over-engineering it yet, but the pattern could outline better once we have one or two more indexers well defined.

3. I think we should try to run a full index to see what ends up in the db and from there decide next step.

faustbrian:
> 1. It's doing exactly what it was doing for elasticsearch but into a sqlite db. But I don't think the data we're persisting is very useful or well digested. For e.g. there's nothing that allows us to query for a transfer transaction. Not sure because I don't know anything about Dot yet but I think we'd need to parse the "method" in the transactions to figure out the transaction type and extract what's useful. I understand you're trying to figure out what's important to index based on what we'll need to query later, so this may get more clear soon.

I am currently looking over 4 different indexers to figure out a common structure we can use because for our use-case we only care about certain values to consume and display. Hope to have something drafted up in the next few days but might take longer to normalise it because there are a lot of differences for even simple things like fees because there are coins like Ethereum which have gas prices and limits stored with a transaction.

> 2. I think there's a pattern that could be extracted for all indexers. For e.g. figure out where to start, fetch blocks + transactions, persist them, repeat until all blocks are consumed. Perhaps we could come up with a template algorithm where each indexer would implement the particulars: fetching transactions (custom for each coin) and requesting storage (elastic search or sqlite for now). Not sure it's worth over-engineering it yet, but the pattern could outline better once we have one or two more indexers well defined.

Once we have all indexers implemented we can take a look at what common patterns we can extract. We can probably define some contracts for things like the database and logging.

> 3. I think we should try to run a full index to see what ends up in the db and from there decide next step.

Will have adrian run a full index the next days to see how it goes now that ethereum works as expected.

PR #1124, opened 2021-03-01
feat(eth-server): pagination for wallet transactions

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-eth-server/src/index.ts
```diff
@@ -126,12 +128,24 @@ export const subscribe = async (flags: {
...
+				params: Joi.object({
+					wallet: Joi.string().length(42),
+				}).options({ stripUnknown: true }),
+				query: Joi.object({
+					page: Joi.number().optional().min(1).default(1),
+				}).options({ stripUnknown: true }),
+			},
+		},
+		handler: (request) => {
+			const offset = (request.query.page - 1) * PAGE_SIZE; // TODO Do we start page numbers at 1 or at 0?
```
marianogoldman, line 142: @faustbrian Pages start at 0 or at 1?

faustbrian, line 142: Pages always start at 1. If we would use the offset as a raw value we would start at 0, so this should be correct.

packages/platform-sdk-eth-server/src/index.ts
```diff
@@ -3,6 +3,8 @@ import Joi from "joi";
 
 import { useClient, useDatabase, useLogger } from "./helpers";
 
+const PAGE_SIZE = 25; // TODO Do we support customizing this by the user? Like a param pageSize or limit?
```
marianogoldman, line 6: Do we parameterize the page size via a `limit` query param? Perhaps checking it's not greater than some default?

faustbrian: We can leave this at a fixed size because those servers are designed to be only consumed by the SDK itself.

PR #1136, opened 2021-03-02
feat(profiles): implement profile importing and exporting

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-profiles/src/profiles/profile.ts
```diff
@@ -55,6 +55,10 @@ export class Profile implements ProfileContract {
 		this.#walletAggregate = new WalletAggregate(this);
 	}
 
+	public clone(): Profile {
```
marianogoldman, line 58: @faustbrian I'm a bit stuck regarding this, and needing some suggestion. In order to export we said "let's remove everything that's not needed and do a dump". I don't think that will work because it will modify the original object. So, for doing that we need to clone it at least. Now, I've been trying different approaches and each has caveats. So let me recap on the options:

1. We clone the profile, we remove stuff on the clone (filter it), we export the clone. Here we can have the ImportExport service which contains that logic.

2. We define an export method in Profile, that takes the export options as a parameter and produces the string filtering whatever is not needed from its internals. But import / export logic is now coupled to the Profile class.

3. We produce an intermediate representation (perhaps the current `toObject()`) and use that for export. For import it's a bit more complex, as we can be not exporting some data, and we need to see how to restore it when importing. For e.g. network information (which we'll presumably get from the environment).

As for approach #1, it's not so easy to clone the Profile, because it's not just copy #data, but also replicating the content of each of the other internal repositories. So it's a deep clone. Is there anything in JS that can help with this?

Approach #2, we have access to the profile internals so we can decide what to export and what not. But it's all entangled in the same class.

Approach #3 has a mix of those: access to the internals for exporting an object and then the service takes care separately of filtering.

I don't feel 100% for any of the 3. So any other ideas are welcome.

faustbrian: I think the current approach is fine with making a new Profile instance and going from there we can add the `toExport` method you mentioned so the profile can figure out what data it needs to provide.

faustbrian:
Took the address list parts from this PR and moved them out so that we can have a fixed structure for all coins that need to generate lists of addresses and also need it for https://github.com/ArkEcosystem/platform-sdk/pull/1149.

- https://github.com/ArkEcosystem/platform-sdk/pull/1150
- https://github.com/ArkEcosystem/platform-sdk/pull/1151

PR #1189, opened 2021-03-10
refactor(intl): append trailing zeros only if value is truthy

dated:
## Summary

The value returned by `Currency#fromString` called with an empty or zero as input currently looks like:

```ts
Currency.fromString("")
// => { display: "", value: "0000000" }

Currency.fromString("0")
// => { display: "0", value: "00000000" }
```

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
Task linked: [CU-ff46zf [Fee] error display when there is no value](https://app.clickup.com/t/ff46zf)

PR #1212, opened 2021-03-16
feat: implement `LedgerService#scan`

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk/src/contracts/coins/ledger.ts
```diff
@@ -21,4 +23,6 @@ export interface LedgerService {
 	signTransaction(path: string, payload: Buffer): Promise<string>;
 
 	signMessage(path: string, payload: Buffer): Promise<string>;
+
+	scan(): Promise<WalletData[]>;
```
marianogoldman, line 27: Just want to be sure this is all we need, or whether we need any more params? Like start scanning from wallet index **n**, etc. Is it ok to return `WalletData[]`? It will be very basic data (like balance), not all data in `WalletData`.

packages/platform-sdk-ark/src/services/ledger.ts
```diff
@@ -41,4 +58,42 @@ export class LedgerService implements Contracts.LedgerService {
...
+		const slip44 = this.#config.get<number>("network.crypto.slip44");
+
+		let wallets: Contracts.WalletData[] = [];
+		let collection: WalletDataCollection;
+		do {
+			const addresses: string[] = [];
+
+			for (const index of createRange(page, pageSize)) {
+				const path: string = formatLedgerDerivationPath({ coinType: slip44, account: index });
+				const publicKey = await this.getPublicKey(path);
```
marianogoldman, line 74: This is actually interacting with the nano ledger. It might eventually have a failure or 2. Do we need a retry policy here?

packages/platform-sdk-ark/src/services/ledger.ts
```diff
@@ -41,4 +58,42 @@ export class LedgerService implements Contracts.LedgerService {
...
+				addresses.push(await this.#identity.address().fromPublicKey(publicKey));
+			}
+
+			collection = await this.#client.wallets({ addresses: addresses });
+			wallets = wallets.concat(collection.items());
+
+			page++;
+		} while (!collection.isEmpty());
```
marianogoldman, line 82: We stop when all the generated addresses are cold wallets. Is this fine? Any other stopping criteria?

packages/platform-sdk-ark/src/services/ledger.ts
```diff
@@ -41,4 +58,42 @@ export class LedgerService implements Contracts.LedgerService {
 	public async signMessage(path: string, payload: Buffer): Promise<string> {
 		return this.#transport.signMessageWithSchnorr(path, payload);
 	}
+
+	public async scan(): Promise<Contracts.WalletData[]> {
+		const pageSize = 5;
```
marianogoldman, line 63: Is chunk size 5 fine? We'll later make wallet requests 5 at a time. Depending on how many at a time the api supports we could have bigger pages.

marianogoldman:
Some of the code here (the iteration for address derivation) will most certainly look the same for every coin. Perhaps the differences will be in regards how to fetch data about the wallets. So some of this could probably belong to some other package and then be reused here.
Perhaps when doing the next coin we can see what it looks like and refactor to extract that common code somewhere common.

packages/platform-sdk-ark/src/services/ledger.ts
faustbrian: With ARK we can send batches of up to 50 addresses in a single request if ever needed.

packages/platform-sdk-ark/src/services/ledger.ts
faustbrian: How often do those errors occur? Don't think any of our products ever had a retry mechanism so far.

packages/platform-sdk-ark/src/services/ledger.ts
faustbrian: Should be fine since that is how the Desktop Wallet works if I remember right.

faustbrian:
Looks good so far but will test it tomorrow when I'm back.

packages/platform-sdk/src/contracts/coins/ledger.ts
faustbrian: For now we shouldn't need any parameters but we'll see what we need once we add more coins and have to handle all their quirks.

PR #1256, opened 2021-03-29
refactor(profiles): split profile restoration into separate methods

faustbrian:


packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.ts
```diff
@@ -274,9 +257,21 @@ export class Profile implements IProfile {
...
+	}
+
+	/**
+	 * Sync the walles and contacts with their respective networks.
+	 *
+	 * @returns {Promise<void>}
+	 * @memberof Profile
+	 */
+	public async sync(): Promise<void> {
+		await this.wallets().syncAll();
```
marianogoldman, line 272: Can these two be done simultaneously? With a Promise.all()?

faustbrian: Missed your comment but no because contacts re-use some data and instances from restored wallets.

PR #1303, opened 2021-04-12
refactor(sdk): determine if coin instance uses a custom host

faustbrian:


packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.ts
```diff
@@ -97,7 +98,16 @@ export class Wallet implements IReadWriteWallet {
...
 			);
 		} else {
-			this.#coin = makeCoin(coin, network);
+			if(container.get<ICoinService>(Identifiers.CoinService).has(coin, network)) {
+				this.#coin = makeCoin(
+					coin,
+					network,
+					{},
+					container.get<ICoinService>(Identifiers.CoinService).get(coin, network).usesCustomHost(),
+				);
```
goga-m, line 107: @faustbrian setting `useForce` explicitly to true, recreates the coin, and updates with new peers. Otherwise, the existing coin with cached peer remains.

```suggestion
			if (container.get<ICoinService>(Identifiers.CoinService).has(coin, network)) {
				this.#coin = makeCoin(coin, network, {}, true);

```

PR #1333, opened 2021-04-20
feat(btc-indexer): store transaction fees

marianogoldman:
## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
Task linked: [CU-h79k1u [BTC] Calculate transaction fee](https://app.clickup.com/t/h79k1u)

packages/platform-sdk-btc-indexer/src/logger.ts
```diff
@@ -0,0 +1,12 @@
+export default class Logger {
```
marianogoldman, line 1: Do we have any other logger better than this humble one?

faustbrian: That should work for now. I'll make one tomorrow that we can share across packages with no issues.

PR #1380, opened 2021-04-29
chore: update coin libraries

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
Polkadot looks like an ESM issue again so maybe we keep it on the current version until that is sorted.

marianogoldman:
Still need to go through more changes, but if we could merge this one that seems ok, so that when I get into issues that require code changes, we can have them isolated.

PR #1382, opened 2021-04-29
refactor(profiles): remove duplicate name type check

w3ea:
## Summary

refactor flush functions in Profile
refactor path of some import module

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
What's the intention of this change? A profile is required to have a name, which should be preserved when you reset the profile settings and data.

w3ea:
> What's the intention of this change?

to remove duplicate codes ([this statement](https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.ts#L226-L228))
we use the name just inside _initialiseSettings_ function and before use it we check name's type ([here](https://github.com/ArkEcosystem/platform-sdk/blob/7575a667ecc1417c4fd2dd39cf363433b4fed4d2/packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.initialiser.ts#L28-L30))

you right about name and i think i didn't touch name

w3ea:
> You changed the name so that it can be `undefined` which isn't the intended behaviour, should always be a string.

so in this case we should remove [this lines](https://github.com/ArkEcosystem/platform-sdk/blob/7575a667ecc1417c4fd2dd39cf363433b4fed4d2/packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.initialiser.ts#L28-L30) because name always is a string, right?

faustbrian:
Correct, a leftover from refactoring. The consumer has to ensure a valid string is passed, thus the 2 guard clauses that you removed.

PR #1388, opened 2021-04-30
chore: update coin libs

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-egld/src/services/transaction.ts
+++ b/packages/platform-sdk-egld/src/services/transaction.ts
@@ -1,16 +1,17 @@
 import { Coins, Contracts, Exceptions } from "@arkecosystem/platform-sdk";
 import { Arr } from "@arkecosystem/platform-sdk-support";
-import { ProxyProvider, UserSigner } from "@elrondnetwork/erdjs";
 import {
 	Address,
 	Balance,
 	GasLimit,
 	GasPrice,
 	Mnemonic,
+	ProxyProvider,
 	Transaction,
 	TransactionPayload,
 	UserSecretKey,
-} from "@elrondnetwork/erdjs/out";
+	UserSigner,
+} from "@elrondnetwork/erdjs";
 
 import { SignedTransactionData } from "../dto";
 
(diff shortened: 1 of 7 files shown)


packages/platform-sdk-egld/src/services/transaction.ts
```diff
@@ -42,18 +52,28 @@ export class TransactionService implements Contracts.TransactionService {
...
+
+		const tx: Transaction = new Transaction(
+			{
+				receiver: Address.fromString(input.data.to),
+				value: Balance.egld(input.data.amount),
+				gasPrice: new GasPrice((input.fee as unknown) as number),
+				gasLimit: new GasLimit((input.feeLimit as unknown) as number),
+				data: new TransactionPayload(input.data.memo),
+			});
+		// tx.setNonce(account.nonce); // TODO Do we need specify this? If so we will need to sync account first, or receive it via param
```
faustbrian, line 73: Unless the library automatically sets the nonce we will need to pass it in.

packages/platform-sdk-egld/src/services/transaction.ts
```diff
@@ -1,8 +1,18 @@
...
+import {
+	Address,
+	Balance,
+	GasLimit,
+	GasPrice,
+	Mnemonic,
+	Transaction,
+	TransactionPayload,
+	UserSecretKey
+} from "@elrondnetwork/erdjs/out";
```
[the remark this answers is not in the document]
marianogoldman: Good question. Checking.

packages/platform-sdk-egld/src/services/transaction.ts
faustbrian: @marianogoldman but we don't need to sync here, it's parameter that can already be passed in.

marianogoldman: I agree... this is their [example code](https://elrondnetwork.github.io/elrond-sdk-docs/erdjs/latest/#creating-value-transfer-transactions) for signing transaction.

I think we need to pass it in, as I don't think the library has the means (at this stage) to figure that out. And yes, I guess it should come as a param in the `TransactionInputs`.

marianogoldman: We don't have any tests that actually ever submit a transaction (like we have for Cardano). Has this ever been tested against the network?

faustbrian: I've manually tested it with the previous lib.

PR #1389, opened 2021-05-02
refactor(profiles): enable password verification for an unrestored profile

goga-m:
## Summary
Because `password.usesPassword()` and `profile.getAttributes().get("password")` as properties can be accessed before a profile is restored, they can also be used to validate a given password of a given profile before a restoration. 

This PR covers the use case where password verification is needed before a profile is restored such as signing in with a password-protected profile in DW.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
Closing in favour of https://github.com/ArkEcosystem/platform-sdk/pull/1390/files

PR #1399, opened 2021-05-05
refactor(profiles): gather all coins while restoring a profile

faustbrian:
> https://app.clickup.com/t/h57fc0

This removes the responsibility of creating instances from all functions. Most functions now expect a coin instance instead of a coin and network name. At the moment this means slightly more work for the client but we should try to either avoid this completely or keep it to a minimum.


packages/platform-sdk-profiles/src/drivers/memory/contacts/contact-address.ts
```diff
@@ -18,6 +18,9 @@ export class ContactAddress implements IContactAddress {
 	}
 
 	public static async make(data: IContactAddressData): Promise<ContactAddress> {
+		// @TODO: we need access to the profile to create a coin instance here but
```
faustbrian, line 21: @marianogoldman any idea for a refactor here to make sure we have an instance passed in but not requiring extra work from the client?

packages/platform-sdk-profiles/src/drivers/memory/repositories/wallet-repository.ts
```diff
@@ -234,7 +242,10 @@ export class WalletRepository implements IWalletRepository {
 
 			wallet.settings().fill(settings);
 
-			await wallet.mutator().coin(coin!, network, { sync: false });
+			await wallet.mutator().coin(
+				new CoinFactory(this.#profile).make(coin, network),
```
faustbrian, line 246: Here we also need a coin instance but we only have the coin and network name at this point.

packages/platform-sdk-profiles/src/drivers/memory/contacts/contact-address.ts
marianogoldman: From the usages (current ones) it seems this is only a helper method you extracted to isolate `ContactAddress` creation, so it won't bother the client, as all they'd do is `contacts().create(...)` on the `ContactAddressRepository`. The repository would be the one to have the coin instance, right? I mean... inside a profile, do we have one contact repo per wallet? one for the whole profile? Or one per coin? If the 1st, we're good, the repo should have the instance and pass it on. If the last one, probably the same. Problem would be if it was a singleton for the profile (which I think it's not).

faustbrian: Contacts are per-profile data, they are not attached to any specific wallet.

packages/platform-sdk-cli/src/commands/import-ledger-wallet.ts
```diff
@@ -132,8 +134,10 @@ export const importLedgerWallet = async (env: Environment, profile: Contracts.IP
 				for (const address of addresses) {
 					imports.push(
 						profile.wallets().push(
+							// @TODO: create coin instance to pass in here
```
faustbrian, line 137: Remember to refactor this.

packages/platform-sdk-profiles/src/drivers/memory/profiles/services/coin.factory.ts
```diff
@@ -0,0 +1,47 @@
...
+			return this.#profile.coins().get(coin, network);
+		}
+
+		const instance = Coins.CoinFactory.make(container.get<Coins.CoinSpec>(Identifiers.Coins)[coin.toUpperCase()], {
+			network,
+			httpClient: container.get(Identifiers.HttpClient),
+			...options,
+		});
+
+		// @TODO: we should move this out but that means potentially more repetitive work for the client
```
faustbrian, line 42: I'm torn on what to do here. This is just a factory and shouldn't store the coin but otherwise the client will have to do more work every time they need to create a coin instance.

I think the best solution here would be to create a separate function or use case that combines the factory and repository so that the client doesn't have to deal with the storage of the coin.

packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.synchroniser.ts
```diff
@@ -12,15 +12,17 @@ export class WalletSynchroniser implements IWalletSynchroniser {
 	/** {@inheritDoc IWalletSynchroniser.coin} */
 	public async coin(options: { resetCoin: boolean } = { resetCoin: false }): Promise<void> {
 		if (options.resetCoin) {
-			this.#wallet.getAttributes().set(
-				"coin",
-				this.#wallet.profile()
-					.coins()
-					.push(this.#wallet.coinId(), this.#wallet.networkId(), {}, true),
-			);
+			// @TODO: create new instance without custom configuration
```
faustbrian, line 15: This is a bit of an odd one because we would need to create a new instance if a reset of the coin is requested but we want to avoid to create coin instances on our own.

faustbrian:
Thinking that one possible solution to the coin creation that is required in a few places is that we could introduce a service that keeps track of all existing coins and networks that exist in wallets and contacts. This would then mean that the client could use that service to construct coins before they are required for an operation - if the coin instance doesn't exist when an operation is executed an exception would be thrown and the client should create the instance.

PR #1409, opened 2021-05-06
docs: sdk usage code samples

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-examples/src/examples/transfer.ts
```diff
@@ -0,0 +1,57 @@
...
+	const contactAddress: Contracts.IContactAddress = await gimli.addresses().first();
+	const destinationAddress: string = contactAddress.address();
+	logger.log("Gimli's Wallet", destinationAddress);
+
+	const transactionId = await wallet
+		.transaction()
+		.signTransfer({
+			from: wallet.address(),
+			sign: {
+				mnemonic: mnemonic1
```
marianogoldman, line 36: @faustbrian Question... given that I'm using a RW wallet from the profile, accessed the profile with the password... is there a way I don't need to provide the mnemonic again for signing this transfer?

marianogoldman: Oh, I thought in DW you didn't need to input that everytime. I don't use DW a lot. And when I did, I used my Nano X, so no need.

faustbrian: Yeah with a ledger you don't need to enter the mnemonic directly but the mnemonic is supposed to be never stored anywhere so every time you want to sign a transaction you need to enter it. Alternatively you can use BIP38 which stores your private key as an encrypted WIF with a custom password which is faster to type.

PR #1445, opened 2021-05-10
fix(profiles): make password available when restoring profile

goga-m:
## Summary
When restoring a password-protected profile (unpack), make the password available from password manager throughout profile's session for later use (e.g auto-saving in emitter)

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/drivers/memory/index.ts
+++ b/packages/platform-sdk-profiles/src/drivers/memory/index.ts
@@ -52,9 +52,11 @@ export class MemoryDriver implements Driver {
 			try {
 				const profile = container.get<IProfileRepository>(Identifiers.ProfileRepository).findById(id);
 
-				if (profile.password().exists()) {
+				if (profile.usesPassword() && profile.password().exists()) {
 					profile.getAttributes().set("data", new ProfileExporter(profile).export(profile.password().get()));
-				} else {
+				}
+
+				if (!profile.usesPassword()) {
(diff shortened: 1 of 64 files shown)


packages/platform-sdk-profiles/src/drivers/memory/index.ts
```diff
@@ -52,9 +52,11 @@ export class MemoryDriver implements Driver {
 			try {
 				const profile = container.get<IProfileRepository>(Identifiers.ProfileRepository).findById(id);
 
-				if (profile.password().exists()) {
+				if (profile.usesPassword() && profile.password().exists()) {
 					profile.getAttributes().set("data", new ProfileExporter(profile).export(profile.password().get()));
-				} else {
+				}
+
+				if (!profile.usesPassword()) {
```
w3ea, line 59:
```suggestion
				else {
```
i think here you can use else condition

goga-m: @w3ea I intentionally added this check so that it would only save the profile if this condition is met. For any other condition, most probably there's something wrong with this profile and we wouldn't want to save it because it might lead to unexpected behaviours and bugs :hmhm:

w3ea: I also believe flat is always better
it wasn't just too clear (from outside)
maybe add some comments make it better, for 6 months later :)

this is just suggestion, do everything that is better for product :oleglala:

PR #1493, opened 2021-05-13
feat(profiles): implement profile status service

goga-m:
## Summary
Initial implementation of profile status service that keeps and updates profile restoration status. This will allow consumers to determine whether a profile has been restored instead of keeping this state outside of sdk. 

Could also be used internally for various actions (e.g emitting change events only when a profile has been restored)

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.status.test.ts
```diff
@@ -0,0 +1,28 @@
...
+	profile = new Profile({ id: "id", name: "name", avatar: "avatar", data: Base64.encode("{}") });
+});
+
+it("should mark as restored", async () => {
+	expect(profile.status().isRestored()).toBeFalse();
+	profile.status().markAsRestored();
+	expect(profile.status().isRestored()).toBeTrue();
+});
+
+it("should reset profile status to default values", async () => {
```
faustbrian, line 23:
```suggestion
it("should reset the status of the profile to the default values", async () => {
```

packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.status.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+let profile: IProfile;
+
+beforeAll(() => bootContainer());
+
+beforeEach(async () => {
+	profile = new Profile({ id: "id", name: "name", avatar: "avatar", data: Base64.encode("{}") });
+});
+
+it("should mark as restored", async () => {
```
faustbrian, line 17:
```suggestion
it("should mark the profile as restored", async () => {
```

PR #1494, opened 2021-05-13
refactor(profiles): save changes only for restored profiles

goga-m:
## Summary
- [x] Reduce profile save calls only when a profile is restored. Change events will be ignored if the profile is not restored as there is no need to save the profile (performance improvement).
- [x] Ensure profile default data is always encoded and saved upon profile creation (don't rely on change emitter).

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/drivers/memory/index.ts
+++ b/packages/platform-sdk-profiles/src/drivers/memory/index.ts
@@ -52,6 +52,10 @@ export class MemoryDriver implements Driver {
 			try {
 				const profile = container.get<IProfileRepository>(Identifiers.ProfileRepository).findById(id);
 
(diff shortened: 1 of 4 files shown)


goga-m:
@faustbrian this PR introduces profile `restore` to be called after profile creation **if we want to save the data** because the profile is not considered restored after creation. Seems more consistent but we could make adjustments on this if you think it's required (will be a breaking a change for some cases in dw)

goga-m:
My assumption was actually wrong. Modified profile creation so that it's marked as restored as all the data is available and ready to use and there is no reason for the additional calling of profile `restore`. Let me know if you see any issues with this.

PR #1520, opened 2021-05-14
feat(nano): nano transactions & broadcasts

axeldelamarre:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-nano/src/services/identity/public-key.ts
```diff
@@ -15,6 +16,6 @@ export class PublicKey implements Contracts.PublicKey {
 	}
 
 	public async fromSecret(secret: string): Promise<string> {
-		throw new Exceptions.NotSupported(this.constructor.name, "fromSecret");
+		return nanocurrency.derivePublicKey(secret);
```
faustbrian, line 19: What does the secret look like?

faustbrian (changes requested):
Please move the addition of the new identity methods into a new PR.

packages/platform-sdk-nano/src/services/identity/public-key.ts
axeldelamarre: I thought this was the private key

packages/platform-sdk-nano/src/services/peer-host.ts
```diff
@@ -0,0 +1,10 @@
+import { Coins } from "@arkecosystem/platform-sdk";
+import { Arr } from "@arkecosystem/platform-sdk-support";
+
+export const getPeerFromConfig = (config: Coins.Config): string => {
```
axeldelamarre, line 4: I see this function repeated in many places, would it make sense to factor it in some library package?

packages/platform-sdk-nano/src/services/identity/public-key.ts
faustbrian: There's a method called `fromPrivateKey` if you want to implement private key derivation.

packages/platform-sdk-nano/src/services/peer-host.ts
faustbrian: https://github.com/ArkEcosystem/platform-sdk/pull/1542

axeldelamarre:
Still missing tests for broadcast because of inability to use a real network

faustbrian:
@axeldelamarre needs updating to use the new signatory class in tests.

axeldelamarre:
This PR is still a draft because the transfer/broadcasting code needs more thorough testing

packages/platform-sdk-nano/src/services/transaction.test.ts
```diff
@@ -0,0 +1,44 @@
...
+	describe("#transfer", () => {
+		it("should sign transaction", async () => {
+			nock("https://proxy.nanos.cc/")
+				.post("/proxy/")
+				.reply(200, require(`${__dirname}/../../test/fixtures/client/account-info.json`));
+
+			const result = await subject.transfer({
+				signatory: new Signatories.Signatory(
+					new Signatories.MnemonicSignatory({
+						signingKey: identity.mnemonic,
```
axeldelamarre, line 29: @faustbrian Can object creation be simplified? Maybe some of the arguments interfaces

faustbrian: All of this is handled by the `SignatoryService` outside of tests.

axeldelamarre: Can MnemonicSignatory inherit from Signatory to avoid double-constructor?

faustbrian: No, it's intentional like this because the implementations for each signatory differ but we don't want to hint 10 different interfaces/classes. It also provides some protection against calling wrong methods and helpers to determine what kind of signing method should be used without having to litter code with `signatory instanceof XXX` assertions.

None of this is of any concern outside of tests because the `SignatoryService` handles this construction, it's not something that will ever be constructed manually outside of tests.

PR #1532, opened 2021-05-17
fix(trx): normalise amounts

faustbrian:


marianogoldman:
@faustbrian Would you add any assertion on some test that checks this? So we don't break it again?

PR #1534, opened 2021-05-17
refactor(profiles): handle profile validations from `IProfileValidator`

goga-m:
## Summary

Extracts profile validation that is used in profile importer to it's own `IProfileValidator` class. The same validation will be used from profile export when saving profile, to prevent data corruption.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/platform-sdk-profiles/src/drivers/memory/profiles/profile.importer.ts
```diff
@@ -78,71 +78,7 @@ export class ProfileImporter implements IProfileImporter {
 	 * @memberof Profile
 	 */
 	private async validate(data: IProfileData): Promise<IProfileData> {
```
faustbrian, line 80: This method can be removed if it's only a one-liner now.

PR #1537, opened 2021-05-17
feat(trx): message signing

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-trx/src/services/message.test.ts
+++ b/packages/platform-sdk-trx/src/services/message.test.ts
@@ -1,0 +1,29 @@
+import "jest-extended";
+
+import { identity } from "../../test/identity";
+import { createConfig } from "../../test/helpers";
+import { MessageService } from "./message";
+
+let subject: MessageService;
+
+beforeEach(
+	async () =>
+		(subject = await MessageService.__construct(createConfig()))
+);
+
+describe("MessageService", () => {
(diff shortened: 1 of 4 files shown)


packages/platform-sdk-trx/src/services/message.test.ts
```diff
@@ -0,0 +1,29 @@
...
+describe("MessageService", () => {
+	it("should sign and verify a message", async () => {
+		const result: any = await subject.sign({
+			message: "Hello World",
+			mnemonic: identity.mnemonic
+		});
+
+		expect(result).toMatchObject({
+			"message": "Hello World",
+			"signatory": "TAq9SwPACv2Ut6YGJK4T8Pw57AGNmFArdP",
```
marianogoldman, line 23: @faustbrian Is it reasonable that we provide the signer's address here rather than the public key? This is different from Ark. The reason I did this is the trx verifier needs an address, and we don't have Address.fromPublicKey() implemented, although I could try to implement it.

faustbrian: Yeah it differs per coin, Bitcoin for example also uses the address, which is why I named the field `signatory` to keep it generic.

faustbrian (approved):

PR #1549, opened 2021-05-18
fix(xlm): broadcast tx handling

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-xlm/src/services/client.ts
+++ b/packages/platform-sdk-xlm/src/services/client.ts
@@ -106,18 +106,16 @@ export class ClientService implements Contracts.ClientService {
 				result.accepted.push(id);
 			} catch (err) {
 				const { extras } = err.response.data;
-				const transactionId: string = extras.envelope_xdr; // todo: get the transaction ID
+				result.rejected.push(transaction.id());
 
-				result.rejected.push(transactionId);
-
-				if (!Array.isArray(result.errors[transactionId])) {
-					result.errors[transactionId] = [];
+				if (!Array.isArray(result.errors[transaction.id()])) {
+					result.errors[transaction.id()] = [];
 				}
 
 				for (const [key, value] of Object.entries(this.#broadcastErrors)) {
 					for (const operation of extras.result_codes.operations) {
 						if (operation.includes(key)) {
-							result.errors[transactionId].push(value);
+							result.errors[transaction.id()].push(value);
 						}
 					}
 				}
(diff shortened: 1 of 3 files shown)


packages/platform-sdk-xlm/src/services/client.ts
```diff
@@ -106,18 +106,16 @@ export class ClientService implements Contracts.ClientService {
 				result.accepted.push(id);
 			} catch (err) {
 				const { extras } = err.response.data;
-				const transactionId: string = extras.envelope_xdr; // todo: get the transaction ID
```
faustbrian, line 109: Did you check if this value and get transaction ID are the same? Some coins require the ID to be modified after broadcasting because the offline ID differs.

marianogoldman: This value is rubish. It doesn't contain the tx id. Here's [an example](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAA9yI2GpXBGeyxcmlQjTXj1WTUyMbWepr30XQXybeP2yAAAAZAADEzQAAAAFAAAAAQAAAAAAAAAAAAAAAF66HwoAAAAAAAAAAQAAAAAAAAABAAAAAHVeJ%2Fac1LQ3rtwfQEBOWtCKLSzZTnBD5iQ6Q%2FdhveHPAAAAAAAAWvMQekAAAAAAAAAAAAGbeP2yAAAAQBiWYaTm27ZWFOCYpIljljVEK24tVewGm2YtcsUVWGgeZ5VrtXzI5dYizy4%2FT1T%2FR1nKb4TItHxj%2B4th0%2F9NmQw%3D&type=TransactionEnvelope&network=test).
I don't even know why the say it's base64 when it can't really be decoded as such. Not sure what their parser does.

marianogoldman: And no... definitely it's NOT their tx id, it's the uuid we use for all txs.

faustbrian: We would need the real ID somehow to show in the UI for explorer linking purposes.

marianogoldman: But in this case, we're on the catch block. So there's no TX id as the tx couldn't be built / signed. In line 104 we do the right thing of replacing the uuid with proper id. Isn't that enough?

PR #1550, opened 2021-05-18
docs(xlm): example transfer

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
All of the examples will probably need adjusting after the signatory changes.

examples/bin/run
```diff
@@ -29,3 +30,7 @@ if (input[0] === "transfer-ark") {
 if (input[0] === "transfer-trx") {
 	(async () => transferTrx())();
 }
+
+if (input[0] === "transfer-xlm") {
+	(async () => transferXlm())();
```
axeldelamarre, line 35: Also possible to wraps the whole file in
```typescript
(async () => {
  //
})();
to avoid repeating the aync

marianogoldman: Awesome! Thanks @axeldelamarre.

marianogoldman:
> All of the examples will probably need adjusting after the signatory changes.

Sure @faustbrian. That's why it would be better if it was a project that had tests and everyone would notice and fix following code changes. Not meant to be a publishable module, but is there any way we can execute these with nock mocks during CI?

faustbrian:
Maybe we could make these commands part of the `platform-sdk-cli` package? Could prefix those commands with `test:`.

faustbrian:
Yeah but that will be easier once we integrate them into a proper package. I can prepare a PR tomorrow.

PR #1571, opened 2021-05-21
refactor: streamline manifest structure

faustbrian:
- https://app.clickup.com/t/k75twg
- https://app.clickup.com/t/k55g62


packages/platform-sdk/src/coins/network.models.ts
```diff
@@ -17,158 +17,160 @@ export type CoinTransactionTypes =
...
+	bip39?: boolean;
+	bip44?: boolean;
+	bip49?: boolean;
+	bip84?: boolean;
 }
 
 export interface NetworkManifest {
-	id: string;
+	uuid: string;
 	type: string;
```
axeldelamarre, line 45: I wouldn't call this uuid, as they have a very specific format and it's not what I see above

packages/platform-sdk/src/coins/network.models.ts
```diff
@@ -17,158 +17,153 @@ export type CoinTransactionTypes =
...
+export type WalletPermission = "read" | "write";
+
+export interface ImportMethod {
+	default: boolean;
+	permissions: WalletPermission[];
 }
 
+export type MnemonicMethod = "bip39" | "bip44" | "bip49" | "bip84";
+
 export interface NetworkManifest {
```
axeldelamarre, line 36: Without context, NetworkManifest can be initially understood as "internet network" (which I actually thought the first time), but I assume it means more like "blockchain network"? I was ok with CoinManifest. Or else BlockchainManifest?

packages/platform-sdk/src/coins/network.models.ts
```diff
@@ -17,158 +17,153 @@ export type CoinTransactionTypes =
...
+		expirationType: ExpirationType;
+		types: CoinTransactionTypes[];
+		fees: {
+			type: FeeType;
+			ticker: string;
 		};
+		memo?: boolean;
+		utxo?: boolean;
+	};
+	import: {
```
axeldelamarre, line 68: import not part of featureFlags?

packages/platform-sdk-ark/src/networks/ark/devnet.ts
```diff
@@ -1,37 +1,69 @@
 import { Coins } from "@arkecosystem/platform-sdk";
 
 const network: Coins.NetworkManifest = {
-	id: "ark.devnet",
+	uuid: "ark.devnet",
```
marianogoldman, line 4: What was wrong with `id`? And this... is definitely not a `uuid`.

packages/platform-sdk-ark/src/networks/compendia/mainnet.ts
```diff
@@ -1,36 +1,65 @@
...
+	},
+	hosts: [
+		{
+			type: "full",
+			host: "https://apis.compendia.org",
 		},
-		expirationType: "height",
+		{
+			type: "explorer",
+			host: "https://bindscan.io/",
```
marianogoldman, line 22: Not related to this change, but a good opportunity to review: all urls that end with `/` get added an extra `/` when we build the final tx or block url in DW. So at least for explorer links we need to remove trailing slashes here.

packages/platform-sdk-eth/src/services/transaction.ts
```diff
@@ -204,7 +204,7 @@ export class TransactionService implements Contracts.TransactionService {
 						},
 					],
 					payable: false,
-					stateMutability: "nonpayable",
+					stateMutpermissions: "nonpayable",
```
marianogoldman, line 207: `stateMutpermissions` => `stateMutPermissions`

packages/platform-sdk-profiles/src/contracts/wallets/wallet.gate.ts
```diff
@@ -14,7 +14,7 @@ export interface IWalletGate {
 	 * @return {boolean}
 	 * @memberof IReadWriteWallet
 	 */
-	allows(ability: string): boolean;
+	allows(permissions: string): boolean;
```
marianogoldman, line 17: Singular? Or should it be an array?

packages/platform-sdk-profiles/src/contracts/wallets/wallet.gate.ts
```diff
@@ -14,7 +14,7 @@ export interface IWalletGate {
...
 
 	/**
 	 * Determine if the given ability should be denied for the current wallet.
```
marianogoldman, line 20: Docs still mention ability instead of permission.

marianogoldman (commented):
Just a few minor comments Brian. Overall it looks like a good refactor.

packages/platform-sdk-ark/src/networks/compendia/mainnet.ts
faustbrian: Yeah that's something I want to address by introducing a function that normalises the URLs. There's also a few that have query parameters which would need to be put at the end.

packages/platform-sdk-ark/src/networks/ark/devnet.ts
faustbrian: A few networks have a literal parameter called network ID so would want to avoid confusion which network ID is meant.

packages/platform-sdk/src/coins/network.models.ts
faustbrian: Feature flags describe the methods that are safe to call without encountering an exception, this is more just information for the client on what is possible with this network.

packages/platform-sdk/src/coins/network.models.ts
faustbrian: https://github.com/ArkEcosystem/platform-sdk/pull/1571#discussion_r636848607

packages/platform-sdk/src/coins/network.models.ts
faustbrian: Those manifests are for blockchain networks, so `NetworkManifest` should make that obvious I would say. I'm not sure how I would end up thinking of a network connection in the context of the whole project 🤔

packages/platform-sdk-ark/src/networks/ark/devnet.ts
marianogoldman: Still think uuid is not a good idea. It's misleading.

faustbrian: Suggest a better name :trollface:

axeldelamarre: I'm also for avoid uuid.

> confusion which network ID

Call it network name?

packages/platform-sdk/src/coins/network.models.ts
axeldelamarre: Almost every package does some form of http/rpc so it's natural to think about network connections. Especially that most people use IDEs and autocompletes of general terms can bring you to wrong places. But anyway it's no big deal

packages/platform-sdk-eth/src/services/transaction.ts
axeldelamarre: I would even go with stateMutabilityPermissions

packages/platform-sdk-ark/src/networks/ark/devnet.ts
faustbrian: It's not a network name, that's `Devnet` in this case. It's a universally unique identifier **(there's no 2 ark.devnet across any implementation)** for the network based on it's coin and name.

packages/platform-sdk/src/coins/network.models.ts
faustbrian: Yeah but if I go by that assumption I would have to rename a lot of things just under the assumption that someone could think about an HTTP connection and I generally avoid binding my naming conventions and development to the assumption that someone will open a project in a software that will make assumptions based on names that are typed out.

axeldelamarre: Ok then it makes sense.
I still think that the word "Blockchain" should have a more central presence in the type system

packages/platform-sdk-eth/src/services/transaction.ts
faustbrian: That's a parameter defined by the web3 package, not us.

PR #1572, opened 2021-05-21
refactor(btc-indexer): use `prisma` as database layer

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-btc-indexer/src/database.ts
```diff
@@ -45,10 +39,11 @@ export class Database {
 
 		logger.debug(`Using [${databaseFile}] as database`);
 
-		this.#database = sqlite3(databaseFile);
-		this.#logger = logger;
+		this.#prisma = new PrismaClient({
+			log: ["query", "info", `warn`, `error`],
```
axeldelamarre, line 43:
```suggestion
			log: ["query", "info", "warn", "error"],
```

packages/platform-sdk-btc-indexer/src/database.ts
```diff
@@ -115,16 +89,27 @@ export class Database {
...
-														VALUES (:hash, :height)`,
-			)
-			.run({
-				hash: block.hash,
-				height: block.height,
-			});
+	private async storeBlock(block): Promise<any> {
+		const transactions: any[] = [];
+		const utxoUpdates: any[] = [];
+		for (const tx of block.tx || []) {
```
axeldelamarre, line 95: transaction instead of tx avoids the abbreviation

packages/platform-sdk-btc-indexer/src/database.ts
```diff
@@ -134,20 +119,25 @@ export class Database {
 	 * @param {*} transaction
 	 * @memberof Database
 	 */
-	private storeTransaction(transaction): void {
+	private async storeTransaction(transaction): Promise<any> {
 		const amount: BigNumber = getAmount(transaction);
 		const vouts: VOut[] = getVOuts(transaction);
 		const vIns = getVIns(transaction);
```
axeldelamarre, line 125: VIn and VOut naming perhaps a big obscure for the uninitiated. I wouldn't hesitate making types & variable names a bit more explicit here

packages/platform-sdk-btc-indexer/src/database.ts
```diff
@@ -57,14 +52,18 @@ export class Database {
...
+				height: true,
+			},
+		});
+
+		if (lastBlockHeight === undefined) {
 			return 1;
 		}
 
-		return lastBlock.height;
+		return (lastBlockHeight["_max"]?.height as number) || 1;
```
axeldelamarre, line 66: Type assertion unneeded here
```suggestion
		return lastBlockHeight["_max"]?.height || 1;
```

axeldelamarre (commented):
Nitpicks

packages/platform-sdk-btc-indexer/src/database.ts
marianogoldman: Not really sure either else to call them or whether to actually do it. That's the name that comes out of the BTC api. So even if unfamiliar to the uninitiated, I can't think of a more accurate name. All open to ideas though...

faustbrian: `inputs` and `outputs` should describe them the same way but would be less technical.

PR #1601, opened 2021-05-26
chore(btc-indexer) more meaningful variable names

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
+++ b/packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
@@ -1,33 +1,33 @@
 import { BigNumber } from "@arkecosystem/utils";
 
-import { VIn, VOut } from "./types";
+import { Input, Output } from "./types";
 
 export const getAmount = (transaction): BigNumber =>
-	transaction.vout.reduce((c: BigNumber, v) => c.plus(Math.round(v.value * 1e8)), BigNumber.ZERO);
+	transaction.vout.reduce((carry: BigNumber, item) => carry.plus(Math.round(item.value * 1e8)), BigNumber.ZERO);
 
-export const getVOuts = (transaction): VOut[] =>
+export const getOutputs = (transaction): Output[] =>
 	transaction.vout
 		.sort((a, b) => a.n - b.n)
-		.map((vOut) => ({
-			idx: vOut.n,
-			addresses: vOut.scriptPubKey?.addresses,
-			amount: Math.round(vOut.value * 1e8),
+		.map((output) => ({
+			idx: output.n,
+			addresses: output.scriptPubKey?.addresses,
+			amount: Math.round(output.value * 1e8),
 		}));
 
-export const getVIns = (transaction): VIn[] =>
+export const getInputs = (transaction): Input[] =>
 	transaction.vin
-		.filter((vin) => vin.txid !== undefined && vin.vout !== undefined)
-		.map((vin) => ({ txid: vin.txid, vout: vin.vout }));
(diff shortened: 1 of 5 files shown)


packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
```diff
@@ -1,33 +1,33 @@
 import { BigNumber } from "@arkecosystem/utils";
 
-import { VIn, VOut } from "./types";
+import { Input, Output } from "./types";
 
 export const getAmount = (transaction): BigNumber =>
-	transaction.vout.reduce((c: BigNumber, v) => c.plus(Math.round(v.value * 1e8)), BigNumber.ZERO);
+	transaction.vout.reduce((carry: BigNumber, item) => carry.plus(Math.round(item.value * 1e8)), BigNumber.ZERO);
```
axeldelamarre, line 6: I am proposing this to improve readability: https://github.com/ArkEcosystem/platform-sdk/pull/1602
Let's wait for @faustbrian's opinion

packages/platform-sdk-btc-indexer/src/types.ts
```diff
@@ -13,13 +13,13 @@ export interface Flags {
 	batchSize: number;
 }
 
-export interface VOut {
+export interface Output {
```
axeldelamarre, line 16: Nice, feels less like cpp code.

Found this [here](https://learnmeabitcoin.com/technical/vout): The “v” in vout stands for vector, because outputs in the bitcoin source code are stored in a vector data structure (which is basically an array)

packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
marianogoldman: Sure. Still we can do this in a separate PR. There's still another instance of the same further down the same file.

faustbrian: Using the `BigNumber#sum` method will add extra processing because we need to call `Math.round(item.value * 1e8)` either before or after instead of directly during the `reduce`.

packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
```diff
@@ -1,33 +1,33 @@
...
 	if (transaction.vin[0].coinbase) {
 		return BigNumber.ZERO;
 	}
 
 	const outputAmount = getAmount(transaction);
 	const inputAmount: BigNumber = transaction.vin
-		.filter((vin) => vin.txid !== undefined && vin.vout !== undefined)
-		.reduce((c: BigNumber, vin) => c.plus(vouts[vin.txid + vin.vout]), BigNumber.ZERO);
+		.filter((input) => input.txid !== undefined && input.vout !== undefined)
+		.reduce((carry: BigNumber, input) => carry.plus(outputs[input.txid + input.vout]), BigNumber.ZERO);
```
faustbrian, line 30:
```suggestion
		.reduce((accumulator: BigNumber, ({ txid, vout })) => accumulator.plus(outputs[txid + vout]), BigNumber.ZERO);
```

packages/platform-sdk-btc-indexer/src/tx-parsing-helpers.ts
faustbrian:
```suggestion
	transaction.vout.reduce((accumulator: BigNumber, ({ value })) => accumulator.plus(Math.round(value * 1e8)), BigNumber.ZERO);
```

PR #1602, opened 2021-05-26
feat(support): add `BigNumber#sum` method

axeldelamarre:
## Summary

Array#reduce is considered harmful when the underlying intent is simply to do a sum or a string concat.

https://betterprogramming.pub/think-again-before-you-use-array-reduce-28f785b5aea9

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-support/src/bignumber.ts
+++ b/packages/platform-sdk-support/src/bignumber.ts
@@ -141,7 +141,7 @@ export class BigNumber {
 	 * @memberof BigNumber
 	 */
 	public static sum(values: NumberLike[]): BigNumber {
-		return values.reduce((accumulator: BigNumber, currentValue: NumberLike) => carry.plus(value), BigNumber.ZERO);
+		return values.reduce((accumulator: BigNumber, currentValue: NumberLike) => accumulator.plus(currentValue), BigNumber.ZERO);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


packages/platform-sdk-support/src/bignumber.ts
```diff
@@ -133,6 +133,17 @@ export class BigNumber {
 		return BigNumber.make(this.#value.multipliedBy(this.toBigNumber(value)));
 	}
 
+	/**
+	 * Returns the sum of the different values.
+	 *
+	 * @param {NumberLike[]} values
+	 * @returns {string}
```
amacar, line 140: This method returns BigNumber.

axeldelamarre: Fixed.. Having to write types *twice* is probably very error prone

packages/platform-sdk-support/src/bignumber.ts
```diff
@@ -133,6 +133,17 @@ export class BigNumber {
...
 
+	/**
+	 * Returns the sum of the different values.
+	 *
+	 * @param {NumberLike[]} values
+	 * @returns {BigNumber}
+	 * @memberof BigNumber
+	 */
+	public static sum(values: NumberLike[]): BigNumber {
+		return values.reduce((carry: BigNumber, value) => carry.plus(value), BigNumber.ZERO);
```
faustbrian, line 144:
```suggestion
		return values.reduce((accumulator: BigNumber, currentValue) => carry.plus(value), BigNumber.ZERO);
```

Also add a type for `currentValue`.

faustbrian (changes requested):

PR #1610, opened 2021-05-27
fix(btc-indexer): prisma client is not a dev dependency

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-btc-indexer/prisma/migrations/20210527154523_initial_migration/migration.sql
```diff
@@ -0,0 +1,42 @@
...
+    "height" INTEGER NOT NULL,
+    "hash" TEXT NOT NULL,
+
+    PRIMARY KEY ("height")
+);
+
+-- CreateTable
+CREATE TABLE "Transaction" (
+    "hash" TEXT NOT NULL,
+    "blockId" INTEGER NOT NULL,
```
faustbrian, line 12: This is the only column that uses camelCase, should be `block_id`.

PR #1620, opened 2021-05-28
feat: implement `WalletDiscoveryService`

faustbrian:
https://app.clickup.com/t/kk8ew9

### DIFF
--- a/packages/platform-sdk-btc/src/services/wallet-discovery.ts
+++ b/packages/platform-sdk-btc/src/services/wallet-discovery.ts
@@ -1,0 +1,29 @@
+import { Coins, Contracts, Services } from "@arkecosystem/platform-sdk";
+import Bitcoin from "bitcore-lib";
+
+import { bip44, bip49, bip84 } from "./identity/utils";
+
+export class WalletDiscoveryService extends Services.AbstractWalletDiscoveryService {
+	readonly #network: Record<string, any>;
+
+	public constructor(config: Coins.Config) {
+		super();
+
+		this.#network = Bitcoin.Networks[config.get<Coins.NetworkManifest>("network").id.split(".")[1]];
+	}
+
+	public static async __construct(config: Coins.Config): Promise<WalletDiscoveryService> {
+		return new WalletDiscoveryService(config);
+	}
+
+	public async fromMnemonic(
+		mnemonic: string,
+		options?: Contracts.IdentityOptions,
+	): Promise<Contracts.AddressDataTransferObject[]> {
+		return Promise.all([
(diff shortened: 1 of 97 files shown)


packages/platform-sdk-btc/src/services/wallet-discovery.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class WalletDiscovery implements Contracts.WalletDiscoveryService {
+	readonly #network: Record<string, any>;
+
+	public constructor(config: Coins.Config) {
+		this.#network = Bitcoin.Networks[config.get<Coins.NetworkManifest>("network").id.split(".")[1]];
+	}
+
+	public async fromMnemonic(mnemonic: string): Promise<Contracts.AddressDiscoveryDataTransferObject[]> {
+		const [nonSegwit, segwit, nativeSegwit] = await Promise.all([
```
axeldelamarre, line 14: for consistency

```suggestion
		const [legacy, segwit, nativeSegwit] = await Promise.all([
```

packages/platform-sdk-nano/src/services/wallet-discovery.ts
```diff
@@ -0,0 +1,22 @@
+import { Contracts } from "@arkecosystem/platform-sdk";
+
+import { deriveAccount, deriveLegacyAccount } from "./identity/helpers";
+
+export class WalletDiscovery implements Contracts.WalletDiscoveryService {
+	public async fromMnemonic(mnemonic: string): Promise<Contracts.AddressDiscoveryDataTransferObject[]> {
+		const [legacy, standard] = await Promise.all([deriveAccount(mnemonic), deriveLegacyAccount(mnemonic)]);
```
axeldelamarre, line 7: On nault.cc, the bip39 account derivation also takes an account index:

image

Is this handled later? I assume the index is included in the path?

packages/platform-sdk/src/contracts/coins/wallet-discovery.ts
```diff
@@ -0,0 +1,9 @@
+export interface AddressDiscoveryDataTransferObject {
```
axeldelamarre, line 1: This is one of the rare exceptions where I would use the `DTO` acronym instead of `DataTransferObject` across the codebase

packages/platform-sdk-btc/src/services/wallet-discovery.ts
faustbrian: Legacy is another schema that is not yet supported so it will be misleading to name it this.

PR #1621, opened 2021-05-28
refactor: normalize input amounts

axeldelamarre:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
@axeldelamarre conflicts

axeldelamarre:
egld: already normalized, not sure about fee
nano: already in normalized
neo: using neo 2 lib, neo 3 available

faustbrian (changes requested):
Looks good except for changing some constructors from using an object to multiple arguments. The object makes it easier to see what argument is what without having to go check the constructor or rely on an IDE.

axeldelamarre:
> Looks good except for changing some constructors from using an object to multiple arguments. The object makes it easier to see what argument is what without having to go check the constructor or rely on an IDE.

The constructor is private, so always visible and close to __construct, and only has 2 arguments. The object pattern is useful when there is lots of optional arguments, but in this case, it introduces any's which is a big no no in my book, IDE or no.

packages/platform-sdk-eth/src/services/transaction.ts
```diff
@@ -9,44 +9,40 @@ import { IdentityService } from "./identity";
...
-		this.#http = opts.http;
-		this.#peer = opts.peer;
-		this.#chain = opts.network;
-		this.#identity = opts.identity;
-		this.#web3 = new Web3(""); // todo: provide a host?
+		this.#http = config.all().http;
+		this.#chain = config.all().network;
+		this.#identity = identity;
+		this.#peer = Helpers.randomHostFromConfig(config);
+		this.#web3 = new Web3(this.#peer);
```
faustbrian, line 24: This is not a Web3 compatible host, that's why there's still the TODO.

PR #1622, opened 2021-05-28
refactor(config): dryer `randomHostFromConfig`

axeldelamarre:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk/src/helpers.ts
+++ b/packages/platform-sdk/src/helpers.ts
@@ -76,8 +76,11 @@ export const randomHost = (hosts: NetworkHost[], type: NetworkHostType): Network
 export const filterHostsFromConfig = (config: Config, type: NetworkHostType): NetworkHost[] =>
 	filterHosts(config.get<NetworkHost[]>("network.hosts"), type);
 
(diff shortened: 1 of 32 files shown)


packages/platform-sdk/src/helpers.ts
```diff
@@ -76,9 +76,12 @@ export const randomHost = (hosts: NetworkHost[], type: NetworkHostType): Network
 export const filterHostsFromConfig = (config: Config, type: NetworkHostType): NetworkHost[] =>
 	filterHosts(config.get<NetworkHost[]>("network.hosts"), type);
 
-export const randomHostFromConfig = (config: Config, type: NetworkHostType): NetworkHost =>
+export const randomNetworkHostFromConfig = (config: Config, type: NetworkHostType = "full"): NetworkHost =>
 	randomHost(config.get<NetworkHost[]>("network.hosts"), type);
 
+export const randomHostFromConfig = (config: Config, type: NetworkHostType = "full"): string =>
```
axeldelamarre, line 82: I would even rename this to `randomHost`. The fact that it comes from config is always clear from the first argument at call sites

PR #1660, opened 2021-06-01
test(profiles): cover ARK mainnet import

goga-m:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.factory.test.ts
```diff
@@ -102,6 +118,14 @@ test("#fromAddress", async () => {
 
 	expect(wallet.address()).toBe("D61mfSggzbvQgTUe6JhYKH2doHaqJ3Dyib");
 	expect(wallet.publicKey()).toBe("034151a3ec46b5670a682b0a63394f863587d1bc97483b1b6c70eb58e7f0aed192");
+
+	const mainnetWallet = await subject.fromAddress({
```
faustbrian, line 122: Change the block to `describe("#fromAddress", async () => {` and then introduce `test("for mainnet", async () => {` and `test("for devnet", async () => {` so we can mor easily see which one fails if they do fail.

PR #1664, opened 2021-06-01
fix(ada): fix manifest for fee all

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-ada/src/networks/shared.ts
```diff
@@ -32,7 +32,7 @@ export const importMethods: Coins.NetworkManifestImportMethods = {
 
 export const featureFlags: Coins.NetworkManifestFeatureFlags = {
 	Client: ["transactions", "transaction", "wallet", "broadcast"],
-	Fee: ["all"],
+	Fee: [],
```
faustbrian, line 35: This key should be optional so we can remove it if it doesn't support anything.

PR #1666, opened 2021-06-01
refactor(support): respect decimals for denominator in `BigNumber#toHuman`

axeldelamarre:
## Summary

use decimals in BigNumber#toHuman

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk-support/src/bignumber.ts
```diff
@@ -264,26 +276,22 @@ export class BigNumber {
...
+		if (decimals === undefined) {
+			// TODO: remove 8 above and uncomment below once everything has been ported
+			// throw new Error("Specify number of decimals for using toHuman()");
+		}
+		const denominator = this.powerOfTen(decimals);
 		return BigNumber.make(this.#value)
-			.divide(1e8)
+			.divide(denominator)
 			.toFixed(decimals);
 	}
```
axeldelamarre, line 296: Does this function look right? I'm not sure because I struggle to write test cases that make sense

faustbrian: That does look like it should be. What's the issue with writing a test for it?

axeldelamarre: Idk everything I try return the wrong thing. Can you pls give 2-3 test examples using decimals both in make() and toHuman() similar to:
```typescript
	expect(BigNumber.make(123.456 * 1e8).toHuman()).toBe("123.45600000");
	expect(BigNumber.make(123.456789 * 1e8).toHuman()).toBe("123.45678900");
```

faustbrian:
```ts
test.only.each([
	[2, 1e2, "100.00"],
	[3, 1e3, "100.000"],
	[4, 1e4, "100.0000"],
	[5, 1e5, "100.00000"],
	[6, 1e6, "100.000000"],
	[7, 1e7, "100.0000000"],
	[8, 1e8, "100.00000000"],
	[9, 1e9, "100.000000000"],
	[10, 1e10, "100.0000000000"],
])("#toHuman(%s)", (decimals, multiplier, expected) => {
	expect(BigNumber.make(100 * multiplier, decimals).toHuman()).toBe(expected);
});
```

Something like this should probably result in the behaviour you want.

axeldelamarre: Like this error:
image

axeldelamarre: Ok I think I got it

faustbrian: The issue with the screenshot is that you multiplied it by `1e8` instead of `1e6` so calling `toHuman(6)` will result in a larger value since there's a difference of `e2` left.

faustbrian (commented):
Please stick to descriptive commit messages, memes and fun have no place in commit history.

packages/platform-sdk-support/src/bignumber.test.ts
```diff
@@ -94,14 +101,10 @@ test("#isLessThanOrEqualTo", () => {
...
-	expect(BigNumber.make(123.456789 * 1e8).toHuman()).toBe("123.45678900");
-	expect(BigNumber.make(1e8).times(1e8).toHuman()).toBe(`${1e8}.00000000`);
-});
-
-test("#toSatoshi", () => {
-	expect(subject.toSatoshi()).toEqual(BigNumber.make(100000000));
+	expect(BigNumber.make(100 * 1e8).toHuman()).toBe("100");
+	expect(BigNumber.make(123.456 * 1e8).toHuman()).toBe("123.456");
+	expect(BigNumber.make(123.456789 * 1e8).toHuman()).toBe("123.456789");
+	expect(BigNumber.make(1e8).times(1e8).toHuman()).toBe(`${1e8}`);
```
axeldelamarre, line 107: Those expected values seemed more 'human' to me but we can revert to using toFixed() if needed

packages/platform-sdk/src/dto/transaction.ts
```diff
@@ -31,7 +31,11 @@ export abstract class AbstractTransactionData {
 		magistrate: "isMagistrate",
 	};
 
-	public constructor(protected readonly data: KeyValuePair) {}
+	protected readonly decimals?: number;
+
+	public constructor(protected readonly data: KeyValuePair, decimals?: number | string) {
+		this.decimals = typeof decimals === "string" ? parseInt(decimals) : decimals;
+	}
```
axeldelamarre, line 38: I realize introducing a second argument makes things a little hairy here. Calling code and helpers become more complicated. I may switch this to a setter method for this

packages/platform-sdk/src/dto/signed-transaction.ts
```diff
@@ -4,12 +4,17 @@ import { BigNumber } from "@arkecosystem/platform-sdk-support";
...
 export abstract class AbstractSignedTransactionData {
+	protected readonly decimals?: number;
+
 	public constructor(
 		protected identifier: string,
 		protected readonly signedData: RawTransactionData,
 		// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 		protected readonly broadcastData: any,
-	) {}
+		decimals?: number | string,
```
axeldelamarre, line 14: Have an extra constructor argument here seems reasonable

packages/platform-sdk-ada/src/services/client.ts
```diff
@@ -41,7 +43,7 @@ export class ClientService extends Services.AbstractClientService {
 		);
 
 		return Helpers.createTransactionDataCollectionWithType(
-			transactions,
+			transactions.map(transaction => [transaction, this.#decimals]),
```
axeldelamarre: This is awkward

packages/platform-sdk/src/dto/transaction.ts
faustbrian: Think an argument is fine for now. I'm currently reviewing classes like this and thinking about how they could be cleaned up and attach them to a factory.

packages/platform-sdk-ada/src/services/client.ts
faustbrian: See https://github.com/ArkEcosystem/platform-sdk/pull/1666#discussion_r643977704

PR #1713, opened 2021-06-03
feat(profiles): use decimal-aware `BigNumber`s in wallet

axeldelamarre:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.ts
+++ b/packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.ts
@@ -54,6 +54,8 @@ export class Wallet implements IReadWriteWallet {
 	readonly #transactionIndex: ITransactionIndex;
 	readonly #walletImportFormat: IWalletImportFormat;
 	readonly #multiSignature: IMultiSignature;
+	readonly #decimals = 8; // TODO: use value from manifest
+	readonly #denomination = BigNumber.powerOfTen(this.#decimals);
 
 	public constructor(id: string, initialState: any, profile: IProfile) {
 		this.#profile = profile;
@@ -142,7 +144,7 @@ export class Wallet implements IReadWriteWallet {
 	public balance(): BigNumber {
 		const value: Contracts.WalletBalance | undefined = this.data().get(WalletData.Balance);
 
-		return BigNumber.make(value?.available || 0);
+		return BigNumber.make(value?.available || 0, this.#decimals);
 	}
 
 	/** {@inheritDoc IReadWriteWallet.convertedBalance} */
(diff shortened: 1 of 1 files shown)


packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.ts
```diff
@@ -153,7 +155,7 @@ export class Wallet implements IReadWriteWallet {
 
 		return container
 			.get<IExchangeRateService>(Identifiers.ExchangeRateService)
-			.exchange(this.currency(), this.exchangeCurrency(), DateTime.make(), this.balance().divide(1e8));
+			.exchange(this.currency(), this.exchangeCurrency(), DateTime.make(), this.balance().divide(this.#denomination));
```
faustbrian, line 158:
```suggestion
			.exchange(this.currency(), this.exchangeCurrency(), DateTime.make(), this.balance().toHuman());
```

Shouldn't this yield the same?

axeldelamarre: No the method return a `BigNumber`, toHuman is a `string`

faustbrian: Would it be cleaner to introduce a new method on `BigNumber` that returns what `toString()` does but as a `BigNumber`? Then we don't need to manually call this `divide`.

axeldelamarre: Yea good point. Let me check all places where it could be used

packages/platform-sdk-profiles/src/drivers/memory/wallets/wallet.ts
```diff
@@ -54,8 +54,7 @@ export class Wallet implements IReadWriteWallet {
 	readonly #transactionIndex: ITransactionIndex;
 	readonly #walletImportFormat: IWalletImportFormat;
 	readonly #multiSignature: IMultiSignature;
-	readonly #decimals = 8; // TODO: use value from manifest
-	readonly #denomination = BigNumber.powerOfTen(this.#decimals);
+	readonly #decimals = 8; // TODO: use value from manifest?
```
faustbrian, line 57: As the comment says, we should grab this value from the network manifest which should be accessible from within this instance.

axeldelamarre: Am I supposed to pass the config all the way down to the `WalletFactory`?

faustbrian: Not sure why you would need to pass it to the wallet factory? You should be able to do `this.#decimals = this.manifest().get<number>("network.currency.decimals");` at some point in the lifecycle. I'll take a look tomorrow morning if it's not sorted by then.

axeldelamarre: I have this error when I try to call `this.manifest()` in `Wallet#constructor` 

image

faustbrian: The manifests only is accessible after a coin has been set.

PR #1714, opened 2021-06-03
feat(lisk): derive addresses and public keys from ledger

marianogoldman:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/platform-sdk-lsk/src/services/ledger.ts
+++ b/packages/platform-sdk-lsk/src/services/ledger.ts
@@ -36,7 +36,6 @@ export class LedgerService extends Services.AbstractLedgerService {
 
 	public async connect(transport: Services.LedgerTransport): Promise<void> {
 		this.#ledger = await transport.open();
-		// @ts-ignore
 		this.#transport = new DposLedger(new CommHandler(this.#ledger));
 	}
 
@@ -52,6 +51,7 @@ export class LedgerService extends Services.AbstractLedgerService {
 
 	public async getPublicKey(path: string): Promise<string> {
 		const { publicKey } = await this.getPublicKeyAndAddress(path);
+
 		return publicKey;
 	}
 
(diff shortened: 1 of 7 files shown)


packages/platform-sdk-lsk/src/services/ledger.ts
```diff
@@ -80,26 +79,21 @@ export class LedgerService extends Services.AbstractLedgerService {
...
 			const addressIndex = initialAddressIndex + addressIndexIterator;
-			const publicKey: string = HDKey.fromCompressedPublicKey(compressedPublicKey)
-				.derive(`m/0/${addressIndex}`)
-				.publicKey.toString("hex");
-
-			const { address } = await this.#identity.address().fromPublicKey(publicKey);
 
+			const path = `m/44'/${slip44}'/0'/0/${addressIndex}`;
+			const { publicKey, address } = await this.getPublicKeyAndAddress(path);
+			console.log("path", path, "publicKey", publicKey, "address", address);
```
marianogoldman, line 93: @faustbrian I finally got this to work. But apparently LSK, just like ARK, doesn't have multiple addresses for same account. So in this code, we're hard-coding account id to `0` (which we should probably allow for iterating over it to have more wallets, but this is something to resolve with all coins) and then even if we iterate over the address index, nano is returning always the same pubkey and address. This is the sample output.
```
    path m/44'/134'/0'/0/0 publicKey d48522677df50defd175c85072309c7643dbc6bdc63c7665a302579ed2ccaedb address 7399986239080551550L
    path m/44'/134'/0'/0/1 publicKey d48522677df50defd175c85072309c7643dbc6bdc63c7665a302579ed2ccaedb address 7399986239080551550L
    path m/44'/134'/0'/0/2 publicKey d48522677df50defd175c85072309c7643dbc6bdc63c7665a302579ed2ccaedb address 7399986239080551550L
    path m/44'/134'/0'/0/3 publicKey d48522677df50defd175c85072309c7643dbc6bdc63c7665a302579ed2ccaedb address 7399986239080551550L
    path m/44'/134'/0'/0/4 publicKey d48522677df50defd175c85072309c7643dbc6bdc63c7665a302579ed2ccaedb address 7399986239080551550L
```

faustbrian: Looking at `getPublicKeyAndAddress` and `getLedgerAccount` it always only derives based on the account but ignores the change and address.

faustbrian: Looking at how they derive wallets in https://github.com/LiskHQ/lisk-desktop I think we need to do 1 account = 1 wallet.

faustbrian:
Merging this for now to test the first batch of changes in DW.

PR #1719, opened 2021-06-04
refactor(sdk): implement `bigNumber` helper for coin-specific instantiation

faustbrian:
> Continuation of https://github.com/ArkEcosystem/platform-sdk/pull/1718

I'm not super happy with the duplication of the `bigNumber` function that will eventually happen but there isn't really a way around it unless we introduce a version of it that takes a container as a second argument, which will make the DX worse and again result in the same duplication in dozens of files instead of 1 file per coin package.

Thus my compromise is `export const bigNumber = (value: NumberLike): BigNumber => Helpers.bigNumber(value, container);` in each coin package which is basically a decorator. This can later be improved, after https://app.clickup.com/t/kz5qmf has been taken care of.

### DIFF
--- a/packages/platform-sdk-ark/src/dto/transaction.ts
+++ b/packages/platform-sdk-ark/src/dto/transaction.ts
@@ -2,6 +2,8 @@ import { Contracts, DTO } from "@arkecosystem/platform-sdk";
 import { DateTime } from "@arkecosystem/platform-sdk-intl";
 import { BigNumber } from "@arkecosystem/platform-sdk-support";
 
+import { bigNumber } from "../container";
+
 export class TransactionData extends DTO.AbstractTransactionData implements Contracts.TransactionData {
 	public id(): string {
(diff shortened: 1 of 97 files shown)


packages/platform-sdk-ark/src/dto/transaction.ts
```diff
@@ -16,7 +18,7 @@ export class TransactionData extends DTO.AbstractTransactionData implements Cont
 	}
 
 	public confirmations(): BigNumber {
-		return BigNumber.make(this.data.confirmations);
+		return bigNumber(this.data.confirmations);
```
axeldelamarre, line 21: confirmations don't have decimals, but amounts do. how do we differentiate the two?

faustbrian: I did a mass search-and-replace, fixing those at the moment.

PR #1768, opened 2021-06-08
refactor(btc-indexer): store pending blocks in database

faustbrian:
1. Downloaded blocks are stored in the database as pending blocks
2. All blocks between the `localHeight` and `remoteHeight` are downloaded and stored independently from the persistence into the `blocks` and `transactions` table
3. Every 30 seconds we process all pending blocks

The flow now is `download new blocks > store as pending > process pending every 30 seconds > delete processed blocks` which should make things more resilient against crashes because we don't need to re-download a bunch of blocks and we can also have better error handling because we have the data sitting in our database until we successfully process and persist it.

The potential overlap in processing caused by `setInterval` shouldn't be an issue because we `upsert` blocks which means we would be just updating an existing record instead of creating another or causing a collision. I've tested this locally up to 150k blocks and there didn't seem to be any issues so far but there might be some issues revealed when the sync goes up to 550k+ blocks where the processing can take longer but I suspect that `upsert` should take care of avoiding any collisions that would cause a crash.

Also something to keep in mind for possible sequential processing or overlapping issues is that we are now storing blocks in the database and get them sorted by their height in ascending order so we already have a guarantee that blocks are sorted in the right order before we start any processing. I haven't yet decided what to do when a block fails to be persisted, most likely we'll need to guess what the issue is and try to fix it.

### DIFF
--- a/packages/platform-sdk-btc-indexer/bin/meow.js
+++ b/packages/platform-sdk-btc-indexer/bin/meow.js
@@ -22,7 +22,7 @@ module.exports = meow(
 			},
 			batchSize: {
 				type: "number",
-				default: 50,
+				default: 100,
 			},
 		},
 	},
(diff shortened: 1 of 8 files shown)


faustbrian:
Merging for now to test the initial changes.

PR #1769, opened 2021-06-08
chore(lsk): update dependencies

faustbrian:
Still has a bunch of issues and documentation isn't very clear on some aspects. Also need to test if all of those changes actually already work on mainnet since this contains a lot of changes which I don't think are live.


faustbrian:
Closing this for now because there is no estimate on when their stuff will hit mainnet.

PR #1783, opened 2021-06-09
refactor: provide default implementations for DTOs

faustbrian:
Provide a default service for DTOs that works based on exports to avoid duplication. Coins can still provide their own implementation if necessary.


faustbrian:
Task linked: [CU-m3cq5m Provide default implementations for DTOs](https://app.clickup.com/t/m3cq5m)

PR #1784, opened 2021-06-09
chore(ark): use flat directory structure with types as suffix

faustbrian:
> This proposal is based on how Angular handles their naming conventions. See https://angular.io/guide/styleguide#separate-file-names-with-dots-and-dashes for more details on that before taking a look at this PR.

> https://github.com/ArkEcosystem/platform-sdk/tree/alternative-structure/packages/platform-sdk-ark/src

I usually work on multiple packages at once and heavily rely on the search of my editor to jump between files which is currently rather tedious because there's multiple files which could match a term because there can be a transaction service, DTO and/or contract.

Outside of that nuisance there's also the issue that navigating through all the nested folders is getting annoying if you do it over and over again while working on multiple packages so this PR proposes to use a flat project structure by using suffixes to identify file types.

### Structure

- `{NAME}.{TYPE}.ts`
- `{NAME}.{TYPE}.test.ts`
- `{NAME}.{TYPE}.helper.ts`
- `{NAME}.{TYPE}.contract.ts`

### Types

- `test` will indicate that the file contains all tests that ensure the unit behaves as expected.
- `service` will indicate that the file contains a service of some kind, like a ledger service that manages all interactions with a hardware wallet.
- `dto` will indicate that the file contains a data transfer object which is responsible for providing normalised data structures.
- `contract` will indicate that the file contains interfaces and types which serve as implementation contracts.

### Notes

- This just a preliminary list of types which is based on types that are commonly showing up. Types can be added, removed or modified as needed.
- `contract` could be split into `interfaces` and `types` but having a single file for those makes it (in my opinion) to quickly skim over them and imports also easier to track.
- This structure could at first glance look off-putting but once in active use it feels more natural, to me personally at least. Instead of thinking `i want the service that handles transactions` and then typing `ark/services/transaction` you can think `i want the transaction service` and then type `ark/transaction.service` which seems like a minor difference but speeds up things for me personally.

> Everything in this PR is fairly subjective based on me always working on loads of different parts of the SDK so your experience might vary if you have a different workflow or generally work on smaller bits.


marianogoldman:
I like this structure more than what we have now. I'm not so used to doing so massive changes, but still navigating through the files will be easier on this flat structure than having to expand, and expand, and expand one more (like identity service for e.g.).

axeldelamarre:
Agreed, "flat is better than nested".

I'm just not a big fan of the helpers. Good code bases should never use names like "helpers", "tools", "utils". Those are very generic names that don't communicate anything useful and encourage poor separation of concerns. If we want to follow Angular, Angular calls those regular functions "services"

axeldelamarre:
Helper classes are evil: https://saintgimp.org/2009/05/05/helper-classes-are-evil/

I actually thought about making a proposal to completely remove helpers from the SDK. Domain logic / business routines don't belong in something called "helper".

I can make another PR if it makes sense with you

faustbrian:
> If we want to follow Angular, Angular calls those regular functions "services".

Don't want to follow Angular by the book, just use it as an example for a potential structure of our own but they have some good organisational conventions like the naming conventions but the `service` one is an odd example because those generally aren't just collections of random functions in Angular but serve a specific purpose while we have a lot of functions that just serve one specific purpose and are unrelated to others so stuffing them into a service/class seems rather overkill, especially with a language like JavaScript which has been traditionally functional until very recently.

> Helper classes are evil: [saintgimp.org/2009/05/05/helper-classes-are-evil](https://saintgimp.org/2009/05/05/helper-classes-are-evil/)

Don't fully agree with that article because it's something that requires context on a product-by-product basis, so parts of it just don't make sense - especially in 2021 when a lot of things and paradigms have changed or been introduced. Not a huge fan of those black & white blog posts that don't provide any real world examples but instead make up some theoretical scenarios.

The examples in that article are the common "worst of the worst" examples that blog posts like that like to use but will rarely be encountered in reality because it's mostly a look at the strict technical aspect of not violating certain principles, which are fine to violate if warranted, because not every function needs to be forced into a use-case or class. Especially with JavaScript a mix of FP and OOP won't cause any harm because everything is still fully testable.

> I actually thought about making a proposal to completely remove helpers from the SDK. Domain logic / business routines don't belong in something called "helper".

Can submit a PR but start with a very small one that isn't time consuming to illustrate the potential changes. If introduced, those should have a separate type from `service` to make it clear what things are services that are exposed by a client and what are internal services. If they use the `service` type we should store them in an `internal` folder or alike.

**TLDR:** I'm open to reworking the helpers into something else but I don't want to end up with a dozen classes that all have 1-2 functions and are only used internally because that has no advantage over functions that are clearly state their intent just by their name.

axeldelamarre:
You make good points, I wasn't advocating too zealously. I was drawing this inspiration from [DDD](https://en.wikipedia.org/wiki/Domain-driven_design) which brings some good timeless principles (like separating domain stuff from tech stuff).

I would do some simple renamings like `foo.helpers.ts` -> `foo.domain.ts` where appropriate.
In other cases maybe `.helpers` -> `.factory` or `.connector` etc

faustbrian:
Factories are definitely that could be created in a lot of places where plain functions are used at the moment. Rest of the changes also sounds fine to me so will see how that looks when you submit a PR.

PR #1826, opened 2021-06-14
refactor: replace uses of `Helpers#toRawUnit` by `BigNumber#toSatoshi`

axeldelamarre:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/platform-sdk/source/services/transaction.service.ts
```diff
@@ -86,4 +87,8 @@ export class AbstractTransactionService implements Contract {
 	public async estimateExpiration(value?: string): Promise<string | undefined> {
 		return undefined;
 	}
+
+	protected toSatoshi = (value: NumberLike) => {
+		return BigNumber.make(value).toSatoshi(this.configRepository.get<number>("network.currency.decimals"));
```
faustbrian, line 92: Inject the `BigNumberService` instead of constructing the same behaviour manually.

packages/platform-sdk/source/services/transaction.service.ts
```diff
@@ -1,6 +1,8 @@
 /* istanbul ignore file */
 
 import { HttpClient } from "@arkecosystem/platform-sdk-http";
+import { NumberLike } from "@arkecosystem/platform-sdk-support";
+import { BigNumberService } from ".";
```
faustbrian, line 5:
```suggestion
import { BigNumberService } from "./big-number.service";
```

axeldelamarre:
Looks like prettier doesn't like the override keyword

